PR #42, opened 2024-05-13
Add latex support

dexter2206:


src/bartiq/integrations/latex.py
```diff
@@ -0,0 +1,173 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Utilities for rendering estimators in LaTeX."""
+from sympy import latex, symbols
+
+from .._routine import Routine
+from ..compilation._utilities import split_equation
```
mstechly, line 18: Those imports are no good, as we're importing from private modules.
I'm wondering whether in such case we should elevate `split_equation` to public API?

tests/test_latex.py
mstechly, line 1: This should be moved to `integrations` directory.

src/bartiq/integrations/latex.py
dexter2206: I would leave it for now, as the *real* issue is that we store local variables as list of strings and not dictionaries. If we could move from this: `["L=a+b", "n=x+y"]` to this: `{"L": "a+b", "n": "x+y"}`, most usages of `split_equation` would vanish anyway.

mstechly (approved):

PR #44, opened 2024-05-15
Add Routine explorer jupyter widget

sammorley-short:
Implements a basic jupyter widget for exploring a `bartiq.Routine`s instance.


sammorley-short:
@dexter2206 @mstechly I think this is good to go modulo a few things:
1. It is downstream from #42, so won't pass tests until that PR does.
2. The estimate explorer doesn't any tests. Do we want to set up some github actions that test the tutorials to ensure they execute with no errors? Otherwise, not sure how best to test rigorously.
3. I haven't updated the tutorial notebooks to include a reference to this. But once they have been updated in #42, we can add it to one or more of them as useful.
4. Right now, this is pretty bare-bones, but if you can take a look and think of any improvements, they'd be more than welcome!

docs/installation.md
```diff
@@ -30,7 +38,12 @@ pip install poetry
 poetry install
 ```
 
-This will create a virtual environment for you and install all developer and docs dependencies within it.
+!!! info
+
+    If you wish to use the package's jupyter integrations, run `poetry install -E jupyter` instead.
```
mstechly, line 43: @dexter2206 as far as I understand `poetry install --with jupyter` is preferable syntax. Could you explain whether that's correct and why actually?

pyproject.toml
```diff
@@ -14,6 +14,14 @@ sympy = "^1.12"
 pyparsing ="~3.1.2"
 qref = "0.1.0"
 
+# A list of all of the optional dependencies, some of which are included in the
+# below `extras`. They can be opted into by apps.
+ipytree = { version = "^0.2.2", optional = true }
+ipywidgets = { version = "^8.1.2", optional = true }
+traitlets = { version = "^5.14.3", optional = true }
+
+[tool.poetry.extras]
```
mstechly, line 23: Seems to me that we would rather use `groups` to handle this.
What is the tradeoff, why would we use one over another @dexter2206 ?

mstechly:
Hey @sammorley-short, I checked in some items in your TODO list above, here's a more comprehensive TODO list ;)

- [ ] 1. Re. tests: I can update it on Tuesday when I pair with @farshadniayeshpour on some CICD improvements. When it comes to more rigorous tests – I don't have a good idea other than "let's make sure it's not failing". @dexter2206?
- [ ] 2. We could also add some bigger routine to latex tests and also "just check whether it doesn't fail when you try rendering the string with sth like `IPython.display.Math`?
- [ ] 3. `explore_routine` doesn't show up in docs, no idea why. @dexter2206 could you take a look?
- [ ] 4. Adding `explore_routine` to `__all__` in `integrations unfortunately forces us to load it, which essentially makes `ipywidgets` & co an obligatory dependency. What would be a good way around it @dexter2206 ? (That's also why tests are failing now)
- [ ] 5. Apparently some of the later changes in #42 made latex not render nicely in the explorer, even though they render nicely with `Math`. See screenshots below.
- [ ] 6. I feel like `represent_routine_in_latex` is a mouthful, how about changing it to `latex_representation(routine)`?
- [ ] 7. See the review comments as well.

pyproject.toml
dexter2206: Extras are the same as with pip, and are pip installable, you can totally do:
```
pip install bartiq[jupyter]
```
On the other hand, groups are dependencies needed for development, and can be installed only using poetry.

The official docs confirm this: https://python-poetry.org/docs/managing-dependencies/

sammorley-short:
@dexter2206 @mstechly 

Ok, so have got the tests passing, but here are the things I couldn't fix:

- [ ] 1. Getting mypy to pass. I thought it was to do with having the jupyter extras installed, but it seems to be something to do with how the extras interact with mypy.
- [ ] 2. I couldn't get the jupyter module added to the API docs. Probably caused by the same issue.

I'm not sure what the best workaround here is, so ideas very welcome!

mstechly:
Why did you close it though 😮 ?

sammorley-short:
> Why did you close it though 😮 ?

Completely unintentionally! Reopened!

mstechly:
@sammorley-short it turns out that what cause the issue with displaying math in widget was "Math Anywhere" chrome extensions ...
Solved!

src/bartiq/integrations/jupyter/routine_explorer.py
```diff
@@ -0,0 +1,82 @@
...
+        node = node or self.root_node
+        self._add_click_event(node)
+        for child_node in node.nodes:
+            self._add_click_events(child_node)
+
+    def _add_click_event(self, node: Node) -> None:
+        node.observe(self.handle_click, "selected")
+
+    def handle_click(self, event: dict) -> None:
+        EVENTS_LOG.append(event)
```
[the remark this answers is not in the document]
mstechly: Fixed in https://github.com/PsiQ/bartiq/pull/44/commits/2b598ff12c1abf97d3cc828fb818b8b1418d7428

dexter2206 (approved):
LGTM!

PR #48, opened 2024-05-24
Update install steps for dev venv to fix version 0.0.0

### LINKED ISSUE #41: Bartiq gets installed with version 0.0.0
When installing bartiq in a fresh venv with `poetry install` it gets installed with the version `0.0.0` (the last message from the install log I get is `Installing the current project: bartiq (0.0.0)`. 

This is most likely due to the setup with dynamic versioning and the fact that `0.0.0` is set as a default version.

CC: @dexter2206

bdg221:
Fixes #41 

In order to ensure the correct version is displayed as the last message in the install log, the poetry-dynamic-versioning plugin needs to be installed prior to `poetry install` being run.

This PR updates the installation page of the documentation to include the plugin installation.

After running `poetry install`, the following is now seen:
`Installing the current project: bartiq (0.1.2.post5.dev0+8356648)`


mstechly:
As it turns out, it got fixed by changes in the release mechanism and currently works properly without the proposed change.
Thanks for the effort though :)

PR #68, opened 2024-06-14
feat: don't add children costs during compilation

### LINKED ISSUE #63: Compilation should not expose children resources by default
Right now when we run `compile_routine`, children resources get propagated to the parent, so we might get resources with keys such as `child_1.t, child_2.t`. 
This is somewhat confusing and might require extra logic in integrations/postprocessing. It's also unnecessary, since this information is already in the `Routine` object anyway, so if someone wants to get it, they can easily do this.

I think we should simply remove it. Perhaps, we could leave a flag that propagates children costs to the parent, but to be honest, I have no idea what was the reasoning for it and it looks like a legacy logic for me that no longer needs to be there.

CC: @dexter2206

mstechly:
Closes #63

### DIFF
--- a/src/bartiq/compilation/_compile.py
+++ b/src/bartiq/compilation/_compile.py
@@ -12,7 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import copy
 from typing import Any, Optional, cast, overload
 
 from .. import Port, Routine
--- a/src/bartiq/compilation/_compile.py
+++ b/src/bartiq/compilation/_compile.py
@@ -682,7 +682,5 @@ def _split_local_path(path: str) -> tuple[str, str]:
 
 def _remove_children_costs(routine: Routine) -> Routine:
     for subroutine in routine.walk():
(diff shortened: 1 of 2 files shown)


src/bartiq/compilation/_compile.py
```diff
@@ -674,3 +678,11 @@ def _split_local_path(path: str) -> tuple[str, str]:
     """Split path into parent path and local name, much like directory path and a file name."""
     *parent_path, name = path.rsplit(".", 1)
     return ("" if parent_path == [] else parent_path[0]), name
+
+
+def _remove_children_costs(routine: Routine) -> Routine:
+    for subroutine in routine.walk():
+        for resource in copy.copy(subroutine.resources):
+            if "." in resource:
+                del subroutine.resources[resource]
```
dexter2206, line 687: **question**: Did you mean:

```suggestion
        subroutine.resources = {name: res for name, res in subroutine.resources.items() if "." not in name}
```
?

dexter2206 (approved):

PR #69, opened 2024-06-18
53 fix handling root name

### LINKED ISSUE #53: Fix handling root name in `_compile_routine`
When refactoring `bartiq` I left the following comment in `_compile`:
```    
# NOTE: Old compilation algorithm was written with the assumption that root is always an empty string.
# The fact that remove the name and then add it back at the end is a hack solution and needs to be fixed in future.
```

To get around it I introduced this logic:

```
    root_name = routine.name
    routine.name = ""
    ...
    compiled_routine_with_funcs.name = root_name
```

We should get rid of this logic and fix the compilation logic  that it does not make any assumptions about naming of root (or any other subroutine).

mstechly:
WiP version of the compilation handling proper non-empty root name.
I didn't have time to clean it up, I want to get rid of `absolute_path_without_root` and just deal with this behavior using `exclude_root_name` flag in `absolute_path`.

Closes #53

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ coverage.xml
 .benchmarks/
 .ipynb_checkpoints/
 .cache
+.vscode
--- a/src/bartiq/_routine.py
+++ b/src/bartiq/_routine.py
@@ -338,7 +338,11 @@ class Routine(BaseModel):
                 raise ValueError("Ancestor not found.") from e
 
     def absolute_path(self, exclude_root_name: bool = False) -> str:
-        """Returns a path from root."""
+        """Returns a path from root.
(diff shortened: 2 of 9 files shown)


mstechly:
I'm not really 100% happy with what I did here, open for counter-opinions.
I changed `absolute_path` from property to a method and added `exclude_root_name:bool=False` to it.
However, in `_compile` it's mostly used with `exclude_root_name=True`, so perhaps it would be a better default.
But then, that's not what I would expect as default from a function with such a name.

If there's some clearly better option, I'm open to it, but otherwise I think it's not terrible as it is in the end 🤷‍♂️

Issue #16, opened 2024-05-03
Support for mathematical constants


mstechly:
Include the ability to support "non-numeric" number values like $\pi$ or $1 / \sqrt(2)$ when performing numeric evaluation.
I think it would be important for certain simplifications in the compilation, where rotation with exactly pi can be replaced by trivial expression. Not supporting it and counting on float representation of pi might lead to some weird "not close enough to pi" issues.

PR #74, opened 2024-06-21
Improve Value Parsing to Take Mathematical Constants and Functions

sitong1011:
### Issue Resolved: #16

Enhancements:

1. Enhanced the parse_value function to handle known constants, and common math operations like sin(), log(), and sqrt(), besides of numbers.
2. Improved error messaging within the parse_value function for better clarity and specificity.
3. Fixed the style.

Dependencies:

No new external dependencies introduced. Utilizes built-in Python libraries `math` and `re`.


docs/index.md
```diff
@@ -27,7 +27,7 @@ from bartiq.integrations import qref_to_bartiq
 with open("alias_sampling_basic.json", "r") as f:
     routine_dict = json.load(f)
 
-uncompiled_routine = qref_to_bartiq(routine_dict)
+uncompiled_routine = qref_to_bartiq(routine_dict)[evaluate_test_data.yaml](..%2Ftests%2Fcompilation%2Fdata%2Fevaluate_test_data.yaml)
```
mstechly, line 30: [issue] I don't think this change was intentional?

src/bartiq/compilation/_utilities.py
```diff
@@ -92,13 +93,69 @@ def get_children_in_walk_order(routine: T) -> list[T]:
...
-            raise ValueError
-        return value
+        if isinstance(value, NUMBER_TYPES):
+            return value
+    except ValueError:
+        pass
+
+    # Check if it is a mathematical function call
+    try:
+        return evaluate_math_function(value_str)
```
mstechly, line 112: It seems to me that this logic shouldn't belong here.
Looking at the docstring, the intention of `parse_value` was to parse expressions representing a single value, and function call is not a single value.

src/bartiq/compilation/types.py
```diff
@@ -12,12 +12,34 @@
...
 
 # Note: when we drop support for 3.9, we will be able to start
 # using the Number union with isinstance checks, thus eliminating
 # the need of separate NUMBER_TYPES
 NUMBER_TYPES = (int, float)
 Number = Union[int, float]
 
+# Define mathematical constants
+Math_constants: Dict[str, float] = {
+    "pi": math.pi,
```
mstechly, line 26: [issue] Actually `math.pi` returns a float value – what we wanted to have is a symbolic representation of `pi`, e.g.: `sympy.pi`.

src/bartiq/compilation/types.py
```diff
@@ -12,12 +12,34 @@
...
+    "e": math.e,
+}
+
+# Define mathematical functions
+Math_functions: Dict[str, Callable[[float], float]] = {
```
mstechly, line 31: [question] Why did you add this? I mathematical functions are dealt with pretty well in the existing symbolic backend?

mstechly (changes requested):
Apart from the issues already raised, there are no new tests added right now.
I think the simplest test would be to define set of expression that get parsed and we can see whether the mathematical symbols get properly parsed. Some ideas:

- `2*pi` -> `6.28` instead of `2*pi`
- `sin(pi)` -> `0`
- `free_symbols("2*pi") == []` (or whathever the syntax is). `pi` should not be treated as a symbol/variable

src/bartiq/compilation/_utilities.py
sitong1011: Removed.

sitong1011:
Although I created the `parse_constant` function to replace mathematical constants like pi in an expression and ignore case sensitivity after using `as_expression`, I found that simply using` sympify `avoids this issue as `sympify` automatically converts mathematical constants in `sympy` to their correct constant values rather than `Symbol` type. 

Interestingly, `as_expression `also uses the same sympify function. I suspect this discrepancy might be related to the output type of `as_expression`.
```
sympify('pi'): <class 'sympy.core.numbers.Pi'> pi
sympy_backend.as_expression('pi'):<class 'sympy.core.symbol.Symbol'>
```
@mstechly @dexter2206

tests/symbolics/test_sympy_backend.py
```diff
@@ -35,6 +35,33 @@ def test_is_sympy_int(expression, expected):
...
+        # Known constants
+        ("pi", pi),
+        ("PI", pi),
+        ("e", E),
+        ("E", E),
+        ("infinity", oo),
+        ("Infinity", oo),
+        # Expressions with constants
+        ("2 * pi", 2 * pi),
+        ("e + 1", E + 1),
```
mstechly, line 50: I'd add some cases where the constants are parts of a more complicated expression – e.g. input to a function `sin(0.25*pi)`

src/bartiq/compilation/_evaluate.py
```diff
@@ -119,7 +119,20 @@ def _parse_assignments(routine: Routine, assignments: list[str]) -> list[Assignm
     # Sort the assignments based on whether they refer to register size variables or not
     parsed_assignments: list[Assignment] = []
     for variable, value_str in assignment_map.items():
-        value = parse_value(value_str)
+        try:
+            value = parse_value(value_str)
+        except BartiqCompilationError:
+            # str to expression
+            expression = sympy_backend.as_expression(value_str)
```
mstechly, line 126: [issue] It should be using `backend` and not `sympy_backend`. The only reason `sympy_backend` is in this file is because we assign it as a default backend.

tests/symbolics/test_sympy_backend.py
```diff
@@ -35,6 +35,33 @@ def test_is_sympy_int(expression, expected):
...
 
+@pytest.mark.parametrize(
+    "expression_str, expected",
+    [
+        # Known constants
+        ("pi", pi),
+        ("PI", pi),
+        ("e", E),
+        ("E", E),
+        ("infinity", oo),
```
mstechly, line 46: Actually, I don't think we should have `infinity` here, as I don't even know in what cases it could arise. Any examples?

src/bartiq/compilation/_evaluate.py
```diff
@@ -119,7 +119,20 @@ def _parse_assignments(routine: Routine, assignments: list[str]) -> list[Assignm
     # Sort the assignments based on whether they refer to register size variables or not
     parsed_assignments: list[Assignment] = []
     for variable, value_str in assignment_map.items():
-        value = parse_value(value_str)
+        try:
+            value = parse_value(value_str)
+        except BartiqCompilationError:
```
mstechly, line 124: Overall this logic looks suspicious to me as there's non-trivial logic in `except` clause, which always raises my suspicion.
Unfortunately I don't have time to look into it right now, leaving this to @dexter2206.

mstechly (changes requested):

mstechly (approved):

PR #92, opened 2024-07-23
Correct Cost Calculation for Uniform State Preparation (USP) in Alias Sampling Tutorials

sitong1011:
## Description

This pull request corrects the T-gates calculation for the Uniform State Preparation (USP) method in the alias sampling tutorials. The previous formula was:

$8*log_2\left(L/k\right)$ 

where $L$ is the state to prepare (referred to as $d$ in Fig. 12). 

According to the paper, this should be:

$8*\left(ceiling\left(\log_2\left(L/2^k\right)\right)-1\right)$

### Context

This correction is needed to align the implementation with the formula provided in the paper. When $L$ is an odd number (e.g., 121),  `compile_routine()` failed due to a sympy error (division by 0). The cause of the bug was the incorrect representation of the cost calculation.

No new dependencies are added with this correction.

- [ ] I have included test cases validating the introduced feature/fix.


docs/tutorials/02_alias_sampling_basic.ipynb
```diff
@@ -356,275 +323,8 @@
...
-      "text/plain": [
-       "<graphviz.graphs.Digraph at 0x107595ff0>"
-      ]
-     },
-     "execution_count": 8,
-     "metadata": {},
-     "output_type": "execute_result"
-    }
-   ],
+   "metadata": {},
```
mstechly, line 327: It seems to me this cell has not been executed and  is missing output?

mstechly (approved):
Overall looks good, but it looks like some cells have not been executed, please do that, commit and then it's good to merge :)

Issue #24, opened 2024-05-03
Add "aggregation" function


mstechly:
It would be useful to have a reliable way to add "aggregated costs" in bartiq.

I imagine the API would look sth like this:

```
aggregation_dict = {"effective_toffs": {"toffs": 1, "gidney_lelbows": 1}, "aggregated_t_gates": {"toffs": 4, "gidney_lelbows": 4, "t": 1, "rotation": 50}
alg_with_extra_resources = add_aggregated_resources(my_algorithm, aggregation_dict)
```

PR #95, opened 2024-07-25
Add aggregation function for resource decomposition

sitong1011:
## Description

This pull request introduces the "aggregation" function to add aggregated costs in qref format. This feature is based on the issue #24 . The function provides a way to aggregate costs and to more accurate estimate resource in quantum algorithms.

### Implemented Solution

The `add_aggregated_resources` function is added to handle the aggregation of costs based on a provided dictionary (`aggregation_dict`). The function expands nested resources and correctly decompose resource list of the given subroutine.

## Changes Made
- Added `_expand_aggregation_dict` function to handle nested resource expansion.
- Added `add_aggregated_resources` function to perform the aggregation based on the provided dictionary.
- Added test cases for those functions.

## Dependencies
No new dependencies were added with this feature.


src/bartiq/transform.py
```diff
@@ -0,0 +1,160 @@
...
+
+
+import copy
+from typing import Any, Dict, Union
+
+import sympy
+from qref import SchemaV1
+
+def add_aggregated_resources(
+    aggregation_dict: Dict[str, Dict[str, Any]], qref_obj: Union[SchemaV1, dict]
```
mstechly, line 23: We should not work directly on a `qref_obj`, rather on the `bartiq` `Routine` object.

src/bartiq/transform.py
```diff
@@ -0,0 +1,160 @@
...
+        subroutine (Dict[str, Any]): The subroutine to which the resources will be added.
+
+    Returns:
+        Dict[str, Any]: The subroutine with aggregated resources.
+    """
+    if "resources" not in subroutine:
+        return subroutine
+
+    # Initialize a dictionary for aggregated resources
+    aggregated_resources = {res["name"]: copy.deepcopy(res) for res in subroutine["resources"]}
```
mstechly, line 63: [comment] In general, if you can avoid using `deepcopy` it's better to avoid it, as using it defensively can lead to big memory and performance issues.

src/bartiq/transform.py
```diff
@@ -0,0 +1,160 @@
...
+    # Iterate over the original resources in the subroutine
+    for resource in subroutine["resources"]:
+        resource_name = resource["name"]
+        resource_type = resource["type"]
+
+        try:
+            resource_expr = sympy.sympify(resource["value"])
```
mstechly, line 70: [issue] Please use backend instead of bare sympy.

mstechly: I guess that's `backend.parse` ?
Or if you start using native `Bartiq` routines, then all the expressions should already be sympy expressions or ints/floats, so you might not even need this step at all.

src/bartiq/transform.py
```diff
@@ -0,0 +1,160 @@
...
+                        "value": multiplier * resource_expr,
+                    }
+            del aggregated_resources[resource_name]
+
+    subroutine["resources"] = list(aggregated_resources.values())
+    return subroutine
+
+def _process_program(expanded_aggregation_dict: Dict[str, Dict[str, Any]], program: Dict[str, Any]) -> Dict[str, Any]:
+    program = _add_aggregated_resources_to_subroutine(expanded_aggregation_dict, program)
+    if "children" in program:
```
mstechly, line 94: This might get simplified if you use `walk` method of `Routine`.

src/bartiq/transform.py
```diff
@@ -0,0 +1,160 @@
...
+    for resource in aggregation_dict.keys():
+        expanded_dict[resource] = _expand_resource(resource, aggregation_dict, set())
+    return expanded_dict
+
+def _expand_resource(resource: str, aggregation_dict: Dict[str, Dict[str, Any]], visited: set) -> Dict[str, Any]:
+    """
+    Recursively expand resource mapping to handle nested resources and detect circular dependencies.
+
+    Args:
+        resource (str): The resource to expand.
```
mstechly, line 120: [style] No need to provide type of the input param here, if you have have typing in the function signature.

tests/test_transform.py
```diff
@@ -0,0 +1,204 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import pytest
+import sympy
+from bartiq.transform import _expand_aggregation_dict, add_aggregated_resources
```
mstechly, line 17: [issue] Please keep only tests of the public functions here.

mstechly (changes requested):

tests/test_transform.py
```diff
@@ -0,0 +1,204 @@
...
+def test_expand_aggregation_dict(aggregation_dict, expected):
+    assert _expand_aggregation_dict(aggregation_dict) == expected
+
+
+@pytest.mark.parametrize(
+    "aggregation_dict, expected",
+    [
+        (
+            {"B": {"C": 4, "D": 5}, "C": {"D": "3*z"}, "A": {"B": 2, "C": "x-y"}},
+            {"B": {"D": "12*z+5"}, "A": {"D": "3*z*(x - y + 8) + 10"}, "C": {"D": "3*z"}},
```
mstechly, line 46: For the case with symbolic please add the example where we decompose rotation gates to T gates using `log(1/eps)` expression.

src/bartiq/transform.py
```diff
@@ -0,0 +1,160 @@
...
+from typing import Any, Dict, Union
+
+import sympy
+from qref import SchemaV1
+
+def add_aggregated_resources(
+    aggregation_dict: Dict[str, Dict[str, Any]], qref_obj: Union[SchemaV1, dict]
+) -> Union[SchemaV1, dict]:
+    """
+    Add aggregated resources to qref program based on the aggregation dictionary.
```
mstechly, line 26: Please add an example of what does the structure of the `aggregation_dict` looks like in the docstring.

src/bartiq/transform.py
dexter2206: It's `backend.as_expression` 🙂

src/bartiq/transform.py
```diff
@@ -0,0 +1,154 @@
...
+    if resource not in aggregation_dict:
+        visited.remove(resource)
+        return {}
+
+    # Sympify the mapping values for the current resource
+    expanded_mapping = {
+        k: backend.parse(v) if not isinstance(v, Number) else k for k, v in aggregation_dict[resource].items()
+    }
+
+    expanded_mapping = {k: sympy.simplify(v) for k, v in aggregation_dict[resource].items()}
```
mstechly, line 133: We should not be using bare `sympy` anywhere here, always do everything symbolic through backend.
Also, as far as I understand, if we drop `simplify`, we'll just end up with slightly worse-looking expressions, so perhaps let's just remove it from here?

src/bartiq/transform.py
```diff
@@ -0,0 +1,154 @@
...
+from bartiq.compilation.types import Number
+from bartiq.symbolics import sympy_backend
+from bartiq.verification import verify_uncompiled_routine
+
+BACKEND = sympy_backend
+
+
+def add_aggregated_resources(aggregation_dict: Dict[str, Dict[str, Any]], routine: Routine, backend=BACKEND) -> Routine:
+    """
+    Add aggregated resources to bartiq routine based on the aggregation dictionary.
```
mstechly, line 31: As mentioned before, please add an example of how the `aggregation_dict` should look like in the docstring 🙏

tests/test_transform.py
```diff
@@ -0,0 +1,171 @@
...
+    ],
+    "resources": [
+        {"name": "arbitrary_z", "type": "additive", "value": "ceil(5*num/2)"},
+    ],
+    "input_params": ["x", "num"],
+    "local_variables": {"n": "x"},
+}
+
+
+def generate_test(subroutine, input_params, linked_params):
```
mstechly, line 52: [suggestion] I'd personally make it a private function.

tests/test_transform.py
```diff
@@ -0,0 +1,171 @@
...
+        ),
+    ],
+)
+def test_add_aggregated_resources(aggregation_dict, generate_test_fn, expected_output):
+    result = add_aggregated_resources(aggregation_dict, generate_test_fn)
+    compare_routines(result, expected_output)
+
+
+# Define the comparison function
+def compare_routines(routine, expected):
```
mstechly, line 163: [suggestion] This one also could be private. Plus, in this case the comment is unnecessary, it's pretty obvious what this function does from its name.

src/bartiq/transform.py
```diff
@@ -0,0 +1,151 @@
...
+    Returns:
+        Dict[str, Dict[str, Any]]: The expanded aggregation dictionary.
+    """
+    if not isinstance(aggregation_dict, dict):
+        raise TypeError("aggregation_dict must be a dictionary.")
+
+    expanded_dict = {}
+    for resource in aggregation_dict.keys():
+        expanded_dict[resource] = _expand_resource(resource, aggregation_dict, set())
+    return expanded_dict
```
dexter2206, line 102: **Issue:** While correct, the recursive approach taken here can quickly blow up the number of calls. Consider the following, very artificial but still moderately sized, `aggregation_dict`:

```python
{
  "a": {
    "b": 2,
    "c": 3
  },
  "b": {
    "c": 10,
    "d": 4
  },
  "c": {
    "d": 5
  },
  "d": {
    "e": 5
  }
}
```
One would expect that the number of `_expand_resource` calls would be roughly equal to number of resources, meanwhile the following sequence of expansion is observed: "a b c d d c d b c d d c d d". This is not surprising, everytime e.g. "c" is needed it is expanded again, even if it was expanded previously. 

This problem might not sound that bad, but we are parsing expressions on each expansion, and parsing expressions might be expensive, hence I'd rather like we address this before merging.

There are two possible solutions:
- Using memoization.
- Topologically sorting the resources to be expanded, so that the more "primitive" resources get expanded first (I'd prefer this one more).

LMK what you think, I'd be happy to pair on this!

dexter2206 (changes requested):

src/bartiq/transform.py
sitong1011: @dexter2206 please check the _topological_sort function in the latest commit.

dexter2206 (approved):

mstechly (approved):

PR #100, opened 2024-08-06
fix: make resource aggregation not delete resources

mstechly:
## Description

I looked once more at `add_aggregated_resources` and I realized that it deletes the "consumed" resource, so I changed it, as I feel it's valuable information that we don't necessarily want to throw away.

@sitong1011 please take a look and let me know if this change makes sense and doesn't accidentally breaks something 😅 


sitong1011:
moved to the new pull request #109

Issue #106, opened 2024-08-07
Add a parameter in aggregation function to decide whether to remove the original resources


sitong1011:
Introduce a parameter in the `add_aggregated_resources` function that allows users to specify whether the original resource should be deleted after aggregation.

### Expected Result:
Given:
```
routine.resources = [ {"name": "rz", "type": "additive", "value": "1"}, {"name": "T_gates", "type": "additive", "value": "5"}]
aggregation_dict = {"rz": {"T_gates": 50}}
```
If remove_original=True:
`aggregated_routine = add_aggregated_resources(routine, aggregation_dict, remove_original=True)
` Then:
```
aggregated_routine.resources = [{"name": "T_gates", "type": "additive", "value": "55"}]
```
If remove_original=False, then:
```
aggregated_routine.resources = [{"name": "rz", "type": "original", "value": "1"}, {"name": "T_gates", "type": "additive", "value": "55"}]
```

### Proposed Changes
- Add a boolean parameter, remove_original, to the `add_aggregated_resources` function. The function signature will be updated as follows:
`def add_aggregated_resources(routine: Routine, aggregation_dict: Dict[str, Dict[str, Any]], backend=BACKEND, remove_original=True)`
- Modify the function logic to conditionally delete the original resources based on the value of remove_original. Additionally, set the type of the decomposed resources as `original` or similar.
- Add test cases to ensure the new parameter works as expected.

PR #109, opened 2024-08-15
fix: Add parameter to decide wheter remove resource in aggregation function 

sitong1011:
## **Description**

This PR solves issue #106 by adding the `remove_decomposed` parameter to the `add_aggregated_resources` function.

### **Context**:
In certain use cases, it is necessary to preserve the original resources after they have been aggregated. Previously, the `add_aggregated_resources` function automatically removed the original resources after aggregation.

### **Implemented Solution**:
- A boolean parameter `remove_decomposed` has been added to the `add_aggregated_resources` function.
  - **If `remove_original=True`** (default): The original resource is removed after aggregation.
  - **If `remove_original=False`**: The original resource is retained, and its type is marked as "other" to distinguish it from other resources (a straightforward approach; however, there may be more optimized solutions.)
- Test cases have been added to validate the correct functioning of this new parameter in both scenarios (`True` and `False`).

### **Dependencies**:
No new dependencies were added in this PR.

## **Please verify that you have completed the following steps**

- [x] I have included test cases validating the introduced feature.

### DIFF
--- a/src/bartiq/transform.py
+++ b/src/bartiq/transform.py
@@ -83,7 +83,6 @@ def _add_aggregated_resources_to_subroutine(
             if remove_decomposed:
                 del aggregated_resources[resource_name]
             else:
-                # type_str = "_".join([f"decomposed_to_{target}" for target in sub_res])
                 aggregated_resources[resource_name].type = "other"
 
     subroutine.resources = aggregated_resources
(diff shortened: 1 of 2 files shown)


src/bartiq/transform.py
```diff
@@ -37,6 +39,10 @@ def add_aggregated_resources(routine: Routine, aggregation_dict: Dict[str, Dict[
                               "arbitrary_z": {"T_gates": "3*log2(1/epsilon) + O(log(log(1/epsilon)))"},
                               ...
                           }
+        remove_decomposed : Whether to remove the decomposed resources from the routine.
```
mstechly, line 42: [nitpick] There's unnecessary space before `:`

PR #115, opened 2024-09-10
feat!: New compilation engine

### LINKED ISSUE #10: Support non-trivial register sizes
Input register sizes must define new input parameters. This means it is fine to define an input register as having size $N$, but it is not fine to try and define the input size as $2N$, or any other function $f(N)$ (where $N$ is some other param, supplied either as an input param, or as an input size param for another input register).

However, one can imagine many cases where one may want to assert that the input register size is some arbitrary function of another input parameter, such as requiring that one input register is twice the size of another.
Currently, this cannot be supported. This is because doing so would introduce a number of non-trivial problems. A loose overview of these are as follows:

1.  If input register sizes can be arbitrary functions (a.k.a are non-trivial), then they effectively represent outputs of the associated function.
2. Then, if an output port is connected to a non-trivial input, then there is no clear way to interpret how to propagate the associated parameters through the function network.
3. In simple cases, such as an input register size of $2N$ connected to an output port with register size $f(a, b)$, it's clear that this fixes $N=f(a,b)/2$, and hence uses of $N$ can be correctly substituted the other expressions of the function.
4. In more complicated cases, such as an output port is $f(a, b)$ and is connected to an input port with non-trivial size $g(c,d)$. This would hence require that $f(a,b)=g(c,d)$. In this case it's clear that there is no simple, general way to propagate the constraints of the output register size further into the function network.
5. Such cases would therefore create a set of additional "constraints", which exist alongside the final compiled function. It is not clear to me how one would systematically resolve such constrains automatically, nor whether sympy support such advanced analytic solving features.
6. If constraints can't be automatically resolved, this would naturally lead to a proliferation of parameters in the function, making the resultant expressions hard to interpret (without manually resolving the constraints first).

These are the issues that come to mind initially, however, I am sure there are others one might need to consider also.
This issue is to highlight this as a known issue in the current implementation, and to provide a space for us to discuss whether this is a feature we need to support in the long-term.

This might be easier to solve if we move to graph based compilation (https://github.com/PsiQ/bartiq/issues/9).

### LINKED ISSUE #14: Deal with passthroughs in compilation
Current compilation engine can't deal with passthroughs. Currently, they're just ignored by the compilation engine. The workaround we have right now is having a precompilation step, but it would be better if passthroughs were natively supported, as it will reduce complexity.

The crux of the issue with passthroughs is that compilation happens bottom-to-top, but if you have a passthrough, then this inverts the logic and is incompatible with the current way bartiq pushes and pulls params around during compilation. So the problem becomes that you won't know what the low-level param should be until later on in the compilation, which would then require one to backtrack through the circuit at a later time.

It might be easier if we implement #9 .

### LINKED ISSUE #19: Add native support for `nlz` function
`nlz` function (number of leading zeros) shows up in various algorithms it makes sense to add it to the list of supported functions in `sympy_interpreter`, which will allow avoiding passing it through `functions_map` parameter.

### LINKED ISSUE #20: Add support for `through` ports
Even though we have introduced a `through` type of port, it's not used anywhere. We should make use of it, as it would:

- simplify logic in many places 
- introduce a more user friendly interface 
- allow for a more rich representation

There are some nuances that need to be taken into account though, as @sammorley-short pointed out:

- it makes an assumption that it's a passthrough
- if it's a container component, you shouldn't be defining the size of your output register anyway (unless it's also a root)
- through should be only available for leaves.

@dexter2206 responded:
I don't think I understand/agree with the above points.
- I think (maybe naively) that it doesn't matter if `through` denotes a passthrough or an operation that maybe is not passthrough, but preserves in/out sizes. And for passthroughs it could possibly make things much simpler. Right now, for passthroughs, we need two ports, a connection between them, and iirc also we introduce a dummy routine to not have to deal with them during compilation.
-  On the other hand, I think that not having passthroughs at all would be nice. Couldn't we just get rid of them altogether? If we have the following passthrough, say `A#out_0 -> B#in_0 -> B#out_0 -> C#in_0`, couldn't we just replace it with `A#out_0 -> C#in_0`? There's probably something that I'm missing though.
- Why shouldn't you define sizes of a container component? If this is the case, than we should make it a hard requirement (meaning: adding validation step saying that if routine `A` has children, it shouldn't contain output port sizes). I always thought that, on a conceptual level, having output port sizes in containers is allowed, and the fact we ignore them is just a limitation of `bartiq`, not a feature. Was i wrong?
- "Through should be only available for leaves" - if we could get rid of passthroughs (see above) than yes, otherwise I think `through` could be used in passthroughs.

@mstechly responded:
- You can't really get rid of passthroughs. In your example B might be acting on the register and have some T-cost associated with it, but perhaps we just didn't bother (or didn't know the implementation yet) to express that.
- I think if we could detect that there's conflict between the size that user declared and the one that we derived, that could be useful information.

### LINKED ISSUE #22: Improvements that will become possible after we drop Python 3.9 support
This is an issue to track what would be possible to improve if we dropped Python 3.9 support.

- [ ] Using `Union`s in `isinstance` checks. Currently we have to use separate definitions for type hints (e.g. `Value = Union[int, float]`) and for actual runtime checks (e.g. `VALUE_TYPES = (int, float)`, followed by `isinstance(x, VALUE_TYPES)`). In Python 3.10+ one could jus write `isinstance(x, Value)`.
- [ ] The following syntax for typing will be allowed: `str | None` rather than `Optional[str]`. 
- [ ] We could remove `sys.setrecursionlimit` in the Double Factorization test.

CC: @dexter2206

### LINKED ISSUE #28: Get rid of parent backlinking
Right now each "subobject" of `Routine` (e.g.: `Connection`, `Port`, `children`) have a link to their parent.
This is an artifact of the old implementation and causes a lot of issues: makes serialization more complicated, might lead to very hard to trace issues with pydantic, etc.

Getting rid of those backlinks would make `Routine` structure much simpler, though introducing this change might be non-trivial, as compilation depends on those backlinks to parents.

CC: @dexter2206

### LINKED ISSUE #45: Make sure `compile_routine` does not change input object
Currently `compile_routine` method modifies "in-place" the input.
We want to make sure it does not happen.

It can be seen in the basic tutorial – after running `compile_routine(uncompiled_routine)` , the `uncompiled_routine` gets a new resources: `T_gates`, which most likely comes from precompilation.

We'd need to ensure that the public methods do not mutating their inputs, as it can lead to very nasty and hard to debug behavior.

### LINKED ISSUE #64: Should we indicating whether a routine is compiled or not
Currently `Routine` can be in two states: uncompiled or compiled, which have some differences:
- uncompiled one has resources which are defined "locally", while compiled has "global" resources – i.e. symbol `x` represents the same variable in all the subroutines
- uncompiled routine needs to have `linked_params` to propagate information about linkage between routines, while compiled routine doesn't need that.

This is confusing, as every time I look at test cases for `evaluate`, I wonder why there are no `linked_params` and how do we know that the symbols mean the same thing, only later to realize that these are already compiled.

Not sure what's a good way of handling that – two that come to mind are:
1. Add a field "compiled": true/false which would indicate what's the state of the routine?
2. Have separate classes for uncompiled and compiled routine?

It's also worth noting that validation rules for compiled and uncompiled routines should be different (https://github.com/PsiQ/bartiq/issues/18).

This is related to https://github.com/PsiQ/bartiq/issues/21.

CC: @dexter2206

### LINKED ISSUE #72: Passthrough port sizes don't get compiled correctly
After running the following piece of code:

```
import json
from bartiq import Routine, compile_routine, evaluate
from bartiq.integrations import qref_to_bartiq

with open("docs/data/alias_sampling_basic.json", "r") as f:
    routine_dict = json.load(f)

uncompiled_routine = qref_to_bartiq(routine_dict)
compiled_routine = compile_routine(uncompiled_routine)
```
And inspecting `swap`:
```
compiled_routine.children["swap"]
```
I get the following result:
Screenshot 2024-06-19 at 7 26 01 PM

These `P_1` and `P_2` are not defined anywhere, they are placeholder variables coming from the passthroughs. At this stage they should have been replaced by real values coming from ports they're connected to, but something went wrong.

This might get resolved while working on #18 , but if not, needs to be fixed after it's resolved.

I was using `bartiq==0.2.0`.

dexter2206:
## Description

This PR introduces new compilation engine, completely replacing the previously used one. 

Among performance improvements, this has the following benefits:

- Passthrough port sizes are compiled correctly (closes #72), and actually no temporary passthrough children need to be introduced (closes #14)
- Different types of objects are used for compiled and uncompiled routines (closes #64)
- Only immutable objects are used (closes #45)
- Internally used structures are singly linked, and therefore there is no parent backlinking (closes #28)
- Through ports are supported natively (closes #20)
- The `nlz` function is natively supported (closes #19)
- Non-trivial register sizes are supported (closes #10), except an edge case for which a separate issue has been created 

In addition, this PR drops support for Python 3.9, which closes #22.


src/bartiq/_routine.py
```diff
@@ -66,380 +45,190 @@ class PortDirection(str, Enum):
...
+        "ports": {port.name: _port_from_qref(port, backend) for port in program.ports},
+        "input_params": tuple(program.input_params),
+        "resources": {resource.name: _resource_from_qref(resource, backend) for resource in program.resources},
+        "connections": {
+            _endpoint_from_qref(conn.source): _endpoint_from_qref(conn.target) for conn in program.connections
+        },
+    }
+
+
+def _rsplit_once(target: str) -> tuple[str, str]:
```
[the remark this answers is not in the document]
dexter2206: Originally this was to satisfy the typechecker (which actually was right). We use output from this function to create tuples of the form (str, str). To see why the typechecker complained, consider this dummy example:

```python
# dummy function accepting pairs of strings
def foo(s: tuple[str, str]) -> None:
    print(s[0], s[1])

# example usage, which is wrong
s = input()
foo(tuple(s.rsplit(".", 1)))
```

The usage shown above is wrong, because `rsplit(".", 1)` can produce a list of length 1. I circumvented it by making a function `_rsplit_once` that ensures that we get two strings from `rsplit`.

Having said that, I just came up with a simpler solution, see [3351962](https://github.com/PsiQ/bartiq/pull/115/commits/3351962fdf453f2b47acc681c861c7013519a4fa).

src/bartiq/_routine.py
```diff
@@ -66,380 +45,190 @@ class PortDirection(str, Enum):
...
+    children: dict[str, Self]
+    ports: dict[str, Port[T]]
+    resources: dict[str, Resource[T]]
+    connections: dict[Endpoint, Endpoint]
+    constraints: Iterable[Constraint[T]] = ()
 
     @property
-    def output_ports(self) -> dict[str, Port]:
-        """Dictionary of output ports of this routine."""
+    def inner_connections(self) -> dict[Endpoint, Endpoint]:
```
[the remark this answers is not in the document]
dexter2206: Correct! Fixed in [6c8a528](https://github.com/PsiQ/bartiq/pull/115/commits/6c8a5288af8628ab15df8ac25e6b1ea6775d7c8f)

src/bartiq/compilation/_evaluate.py
```diff
@@ -95,234 +49,42 @@ def evaluate(routine, assignments, *, backend=sympy_backend, functions_map=None)
         A new estimate with variables assigned to the desired values.
     """
     return _evaluate(
-        routine=routine,
+        compiled_routine=routine,
```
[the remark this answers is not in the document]
dexter2206: You're right, but actually this issue no longer exists and we can do now with `evaluate` only. Good catch! Fixed in [b87a82c](https://github.com/PsiQ/bartiq/pull/115/commits/b87a82ccbea2a86a554dd94441dcca349b1c0796)

src/bartiq/compilation/_evaluate.py
```diff
@@ -95,234 +49,42 @@ def evaluate(routine, assignments, *, backend=sympy_backend, functions_map=None)
...
-    value = assignment.value
-
-    old_function = to_symbolic_function(routine, backend)
-
-    new_function = define_expression_functions(old_function, functions_map)
-    new_function = set_input_port_size_to_constant_value(new_function, register, value, backend)
-
-    update_routine_with_symbolic_function(routine, new_function)
-    return routine
+def _evaluate_internal(
```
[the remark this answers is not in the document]
dexter2206: (see my previous comment first)
The `evaluate` is a this wrapper around `_evaluate_inner`, which parses constants in assignments and validates inputs. `evaluate_inner` is the actual implementation. The distinction is important, because `_evaluate_inner` is called recursively, and we don't want to perform input validation with every call.

src/bartiq/compilation/_compile.py
```diff
@@ -11,715 +11,234 @@
...
-    routine: Routine,
-    *,
-    precompilation_stages: Optional[list[PrecompilationStage]] = None,
-    global_functions: Optional[list[str]] = None,
-    functions_map: Optional[FunctionsMap] = None,
-    skip_verification: bool = False,
-) -> Routine:
-    pass  # pragma: no cover
-
+ParameterTree = dict[str | None, dict[str, TExpr[T]]]
```
mstechly, line 43: Would be nice to have a comment explaining what it represents?

src/bartiq/compilation/_compile.py
```diff
@@ -11,715 +11,234 @@
...
-            new_param = old_param.removeprefix(namespace_prefix)
-            namespace_map[old_param] = new_param
-
-    return rename_variables(function, namespace_map)
+    connections_map = _expand_connections(routine.connections)
 
+    local_variables = _compile_local_variables(routine.local_variables, inputs, backend)
+
+    # Parameter map holds all of the assignments as nested dictionary.
+    # The first level of nensting is the child name (or None for current routine assignments)
```
mstechly, line 183:
```suggestion
    # The first level of nesting is the child name (or None for current routine assignments).
```

src/bartiq/compilation/_compile.py
```diff
@@ -11,715 +11,234 @@
...
-            new_output_symbol = f"#{port.name}"
-            new_outputs[new_output_symbol] = DependentVariable(
-                new_output_symbol,
-                backend.as_expression(port_endpoint.size),
-                backend=backend,
-            )
-    return SymbolicFunction(function.inputs, new_outputs)
-
+    # We start by populating it with freshly compiled local variables and inputs
+    # {None: {a: 2}, b: {foo: 3, #in_1: N}}
```
mstechly, line 188: To be honest,  this example here is somehow more confusing than revealing to me.

src/bartiq/compilation/preprocessing.py
```diff
@@ -0,0 +1,139 @@
...
+def add_default_additive_resources(routine: Routine[T], backend: SymbolicBackend[T]) -> Routine[T]:
+    child_resources_map: defaultdict[str, set[str]] = defaultdict(set)
+
+    for child in routine.children.values():
+        for resource in child.resources.values():
+            if resource.type == ResourceType.additive:
+                child_resources_map[resource.name].add(child.name)
+
+    additional_resources: dict[str, Resource[T]] = {
+        # The default here is to satisfy the typechecker
```
mstechly, line 34: I don't understand this comment.

src/bartiq/compilation/preprocessing.py
```diff
@@ -0,0 +1,139 @@
...
+        )
+        for res_name, children in child_resources_map.items()
+        if res_name not in routine.resources
+    }
+
+    return replace(routine, resources={**routine.resources, **additional_resources})
+
+
+@postorder_transform
+def promote_unlinked_inputs(routine: Routine[T], backend: SymbolicBackend[T]) -> Routine[T]:
```
mstechly, line 51: I don't like the name. How about `extend_by_unlinked_inputs` or sth like that?

src/bartiq/integrations/__init__.py
```diff
@@ -14,12 +14,9 @@
...
-    additional_all = ["explore_routine"]
 except ImportError:
-    additional_all = []
+    pass
 
 from .latex import routine_to_latex
-from .qref import bartiq_to_qref, qref_to_bartiq
 
-__all__ = ["bartiq_to_qref", "qref_to_bartiq", "rotine_to_latex"] + additional_all
+__all__ = ["routine_to_latex", "explore_routine"]
```
mstechly, line 22: Why this change?

src/bartiq/symbolics/sympy_backends.py
```diff
@@ -163,18 +169,31 @@ def define_function(self, expr: T_expr, func_name: str, function: Callable) -> T
...
         """Return a textual representation of given expression."""
+        if isinstance(expr, Number):
+            return str(expr)
         return serialize_expression(expr)
 
+    def compare(self, lhs: TExpr[Expr], rhs: TExpr[Expr]) -> ComparisonResult:
+        difference = self.as_expression(lhs - rhs)
+        if not isinstance(difference, Number):
+            difference = difference.expand()
+        if difference == 0:
```
mstechly, line 190: What if (for whatever reasons) the difference is `0.0` and not `0` 🙃 ?

dexter2206: Although I think in this particular case this cannot happen, I added additional comparison to 0.0 in [0faae4e](https://github.com/PsiQ/bartiq/pull/115/commits/0faae4ed23be183e11a35c0e9b290b925ac43b6d)

tests/conftest.py
```diff
@@ -24,47 +22,3 @@
 def backend():
```
mstechly, line 22: We don't use this fixture much anyway, I think we could delete this file?
Looks like it's used only in `test_evaluate` and `test_preprocessing`.

Or if it stays, please update `test_compile` to also use it, so we're more consistent?

tests/symbolics/test_sympy_backend.py
```diff
@@ -68,20 +68,6 @@ def test_value_of_returns_none_if_numerical_evaluation_is_not_possible():
     assert sympy_backend.value_of(expr) is None
 
 
-def test_attempt_to_rename_builtin_function_fails():
```
mstechly, line 71: Isn't it good that we had those tests?

src/bartiq/integrations/__init__.py
dexter2206: So:
- There are no `bartiq_to_qref` and `qref_to_bartiq` functions anymore.
- There was a typo in "rotine_to_latex"
- Expression of the form `__all__ = a + b` is an incorrect expression for `__all__` variable (see [here](https://microsoft.github.io/pyright/#/typed-libraries?id=library-interface))

Having said that, we can do better, see [66cc331](https://github.com/PsiQ/bartiq/pull/115/commits/66cc3318ad23a7a97201df42489bcd493931f883)

src/bartiq/compilation/preprocessing.py
dexter2206: Well, that's because it should be there anymore 😄 Fixed in [a4d62fd](https://github.com/PsiQ/bartiq/pull/115/commits/a4d62fd9faa37f5d4c3ff401d04f2884227a16f2)

src/bartiq/compilation/_compile.py
dexter2206: I added a comment explaining this data structure in [075ea09](https://github.com/PsiQ/bartiq/pull/115/commits/075ea09b02a6bae5846a26905655e6c39729078d). WDYT?

src/bartiq/compilation/_compile.py
dexter2206: I agree this example is confusing. I removed it in favor of other comment that I added near the definition of `ParameterTree`.

tests/compilation/test_preprocessing.py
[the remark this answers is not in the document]
dexter2206: Some of those will be added when the main is merged, we can add more later.

However, I think we shouldn't obsess about having tests *specifically* for any given feature. The things that matter are: demonstration of the intended use of the public API, and the coverage, in that order. Currently the preprocessing is 100% covered 🙂

src/bartiq/compilation/_compile.py
mstechly: Much better, thanks!

tests/compilation/test_preprocessing.py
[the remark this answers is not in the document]
mstechly: Sounds good.
Not obsessing over it, but I think the role of preprocessing can be somewhat confusing, so it's good to have it for demonstration reason :)

tests/compilation/test_utilities.py
```diff
@@ -14,12 +14,7 @@
 
 import pytest
 
-from bartiq.compilation._utilities import (
-    is_constant_int,
-    is_non_negative_int,
-    is_single_parameter,
-    split_equation,
-)
+from bartiq.compilation._utilities import is_single_parameter
```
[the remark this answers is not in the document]
dexter2206: Actually, looking at the `_utilities` module right now, it only contains this `is_single_parameter` function, which much better fits the `backend` interface. I moved the definition there, implemented it in SymPy backend, and deleted `_utilities` module altogether in [60e98e](https://github.com/PsiQ/bartiq/pull/115/commits/760e98e84662a8f17fec79877589be3a0aa9ed14).

mstechly (approved):

PR #116, opened 2024-09-14
docs: update tutorials to work with the new compilation engine

mstechly:


dexter2206 (approved):

docs/tutorials/01_basic_example.ipynb
```diff
@@ -655,7 +638,7 @@
    "id": "916e166e-3e89-4201-b24e-7b3802a97dae",
    "metadata": {},
    "source": [
-    "Now it would be good to know what is the cost when we subsitute some numbers. We can do this using `evaluate` method. As you can see in the example below, it can either take all the parameters or just some of them."
+    "Now it would be good to know what is the cost when we subsitute some numbers. We can do this using `evaluate` method. As you can see in the example below, it can either take substitute all the parameters or just some of them."
```
dexter2206, line 641: There is something wrong with last sentence, I think we should delete "take".

```suggestion
    "Now it would be good to know what is the cost when we subsitute some numbers. We can do this using `evaluate` method. As you can see in the example below, it can either substitute all the parameters or just some of them."
```

src/bartiq/compilation/_compile.py
```diff
@@ -101,22 +113,40 @@ def _compile_constraint(
...
     backend: SymbolicBackend[T] = sympy_backend,
-    precompilation_stages: Iterable[PreprocessingStage[T]] = DEFAULT_PRECOMPILATION_STAGES,
+    preprocessing_stages: Iterable[PreprocessingStage[T]] = DEFAULT_PREPROCESSING_STAGES,
     skip_verification: bool = False,
 ) -> CompilationResult[T]:
-    if not skip_verification:
+    """Performs symbolic compilation of a given routine.
+
+    Compilation in this context means taking a routine that's defined in terms of variables defined within the scope
+    of a given routine and returning a routinedefined in terms of top-level variables.
```
dexter2206, line 125: I would slightly alter this description to be more concise (and arguably - more understandable). WDYT about the following:

```suggestion
    In this context, compilation means transforming a routine defined in terms of routine-local variables into one defined in terms of global input parameters.
```

src/bartiq/compilation/_compile.py
```diff
@@ -101,22 +113,40 @@ def _compile_constraint(
...
+
+    Args:
+        routine: routine to be compiled.
+        backend: a backend used for manipulating symbolic expressions.
+        preprocessing_stages: functions used for preprocessing of a given routine to make sure it can be correctly
+            compiled by Bartiq.
+        skip_verification: a flag indicating whether verification of the routine should skipped.
```
dexter2206, line 132: Typo
```suggestion
        skip_verification: a flag indicating whether verification of the routine should be skipped.
```

dexter2206 (changes requested):

dexter2206 (approved):

Issue #12, opened 2024-05-03
Add estimate optimizer module


mstechly:
In many algorithms there might be free parameters, with arbitrary costs (e.g. $\lambda$ in QROM).
We should therefore provide users with some function that can automatically set such parameters via optimization.

Assuming `my_routine` has variables: "a, b, c" and cost "x" it could look sth like this:

```
evaluated_routine = evaluate(my_routine, assignments=["a=1", "b=2")
optimize(evaluated_routine, cost_func="x", variable="c")
```

Adding support for multivariate optimization is something we could introduce in future iterations, but I wouldn't worry about it for MVP.

PR #111, opened 2024-08-21
feat: Add Single-Parameter Optimization Module to Bartiq

sitong1011:
### Context:
This PR addresses issue #12 by integrating a basic optimization module into Bartiq. This enhancement enables users to optimize parameters related to resource expressions within Bartiq routines.
### Implemented Solution:
Added a minimize function to allow optimization of resources with respect to a single free parameter.
Users can specify bounds for the parameter and choose the optimizer.

### Dependencies:
Added scipy to the dev dependencies as one backend could be used by minimize function.

## Please verify that you have completed the following steps

- [ ] I have self-reviewed my code.
- [ ] I have included test cases validating introduced feature/fix.
- [ ] I have updated documentation.

PR #118, opened 2024-09-15
feat: Add Single-Parameter Optimization Module to Bartiq

sitong1011:
### Context:
This PR addresses issue #12 by integrating a basic optimization module into Bartiq. This enhancement enables users to optimize parameters related to resource expressions within Bartiq routines.
### Implemented Solution:
Added a minimize function to allow optimization of resources with respect to a single free parameter.
Users can specify bounds for the parameter and choose the optimizer.

### Dependencies:
Added scipy to the extras dependencies as one backend could be used by minimize function.


sitong1011:
https://github.com/PsiQ/bartiq/pull/111#discussion_r1750138293 @mstechly 
I’ve adjusted the optimizer by reducing the default learning rate to 0.000001 and adding a momentum of 0.9 to adjust the influence of previous updates. If you have any additional examples, it would be great to test whether the changes improve its performance.

Additionally, it seems that the overall df_active_volume is an increasing function. If this assumption is correct, then the current test results should now be reasonable.

sitong1011:
Hi @mstechly, I've completed the changes and everything is ready for tests and your review.

.github/workflows/build.yaml
```diff
@@ -26,7 +26,7 @@ jobs:
         with:
           poetry-version: 1.8.3
       - name: Install the packages required
-        run: poetry install --with docs
+        run: poetry install --with docs --extras "optimization"
```
mstechly, line 29: @dexter2206 I don't think this should be here?

mstechly (approved):

PR #124, opened 2024-09-25
fix: Fix LaTeX rendering in docs

dexter2206:
## Description

Rendering LaTeX formulas in docs didn't work properly (as notice by @mstechly ). This PR fixes it.

- N/A I have included test cases validating introduced feature/fix.
- N/A I have updated documentation.

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -77,7 +77,9 @@ markdown_extensions:
   - pymdownx.inlinehilite
   - pymdownx.snippets
   - pymdownx.superfences
-
+extra_javascript:
+  - javascript/config.js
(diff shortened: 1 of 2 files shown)


mkdocs.yml
```diff
@@ -77,7 +77,9 @@ markdown_extensions:
   - pymdownx.inlinehilite
   - pymdownx.snippets
   - pymdownx.superfences
-
+extra_javascript:
```
mstechly, line 80: If there's a specific issue or SO question that we could add here as a reference on why this is needed, it would be helpful for future us.

mstechly (approved):

mkdocs.yml
dexter2206: I just followed the official docs, I added a comment with a link.

PR #133, opened 2024-10-24
fix: Fix nontrivial input port size support

### LINKED ISSUE #127: Nontrivial input port sizes are incorrectly handled if they are functions of other port size
## Describe the bug
Nontrivial input port sizes can be handled incorrectly if they are defined in terms of variables introduced in other input ports. Such port sizes will be ignored and compiled to a size of the upstream port.

## How to Reproduce
Construct a routine with a single child and the following connections: `in_0 -> child.in_0`, `in_1 -> child.in_1`.
Let both the `in_0` and `in_1` have size `N`, let `child.in_0` have size `M` and let `child.in_1` have size `M**2`.

Then, compile such a routine and evaluate it with any value for `N` > 1.

## Expected behavior
An error should be raised, since clearly such an assignment forces `N = N ** 2`, which is false for `N` > 1. 

## Actual behavior
No error is raised. This behavior occurs because the needed constraint is not added.

## Environment:**
- All Python versions and OSes are affected.

dexter2206:
## Description

Fixes #127 by introducing relevant logic for handling non-trivial input port sizes depending on other ports.

### DIFF
--- a/tests/compilation/test_compile.py
+++ b/tests/compilation/test_compile.py
@@ -141,3 +141,31 @@ def test_compile_errors(routine, expected_error, backend):
         compile_routine(
             routine, preprocessing_stages=[introduce_port_variables], backend=backend, skip_verification=True
         )
+
+
+def test_compilation_introduces_constraints_stemming_from_relation_between_port_sizes(backend):
+    routine = RoutineV1(
+        name="root",
+        type="dummy",
+        children=[
+            {
+                "name": "a",
+                "type": "dummy",
+                "ports": [
+                    {"name": "in_0", "direction": "input", "size": "N"},
+                    {"name": "in_1", "direction": "input", "size": "2 ** N"},
+                ],
+            }
+        ],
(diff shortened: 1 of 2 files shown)


tests/compilation/test_compile.py
```diff
@@ -141,3 +141,48 @@ def test_compile_errors(routine, expected_error, backend):
         compile_routine(
             routine, preprocessing_stages=[introduce_port_variables], backend=backend, skip_verification=True
         )
+
+
+def test_compilation_introduces_constraints_stemming_from_relation_between_port_sizes(backend):
```
mstechly, line 146: I think it's also worth adding a `fun(N)` to make sure parsing functions also works in this context?

mstechly (approved):
Looks good, one minor coment!

tests/compilation/test_compile.py
dexter2206: Done!

PR #135, opened 2024-10-24
feat: Enable casting to native types and obtaining backend-native functions

dexter2206:
## Description

While working on several features, we noticed that there are some limitations to the backend interface, namely:

1. It is not possible to obtain native Python objects from expressions. There is a `serialize` method that always returns string, and there is `value_of` method that either returns a number OR `None` if it is not possible to convert an expression to number. There is no single function that would convert expressions representing numbers to numbers, and other expressions to strings.
2. While manipulating symbolic expressions is easy through arithmetic operators, there's not really a way for applying function to symbolic expressions. The current solution involves going through `backend.as_expression`, e.g.: `backend.as_expression(f"sin({backend.serialize(some_expr)})")` which a) is not very readable , b) incurs a performance penalty if `some_expr` is long and complicated.

This PR solves both of this issues by defining `backend.as_native(expr)` and `backend.func(func_name)` methods.

The `backend.func` can be used as follows:
```python
x = .... # Assume we have some expression for x
sin = backend.func("sin")
result = sin(x). # This is of type TExpr[T] (for backend specialized to T
```
The name `func_name` was chosen instead of e.g. `backend.function` so that we have more concise notation while still maintining a reasonable level of readability.

### DIFF
--- a/src/bartiq/_routine.py
+++ b/src/bartiq/_routine.py
@@ -187,13 +187,13 @@ def _endpoint_from_qref(endpoint: str) -> Endpoint:
 def _port_to_qref(port: Port[T], backend: SymbolicBackend[T]) -> PortV1:
     return PortV1(
         name=port.name,
-        size=backend.serialize(port.size),
+        size=backend.as_native(port.size),
         direction=cast(Literal["input", "output", "through"], port.direction),
     )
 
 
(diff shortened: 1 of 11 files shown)


mstechly (approved):
Great job!

PR #136, opened 2024-11-05
feat: add repeated structures

mstechly:
## Description

Adds support for repeated structures and a couple of other minor features.
For tests to pass it would require https://github.com/PsiQ/qref/pull/124 to go in first an QREF to be released.

There's a couple of things left:
- [x] Documentation
- [x] Some minor test cases are missing
- [x] Moving `_evaluate_and_define_functions` in an appropriate place.


tests/compilation/test_repetitions.py
```diff
@@ -0,0 +1,257 @@
...
+        },
+        version="v1",
+    )
+
+
+def _constant_sequence_sum(unit_cost, count, multiplier):
+    sum = 0
+    for _ in range(count):
+        sum += unit_cost * multiplier
+    return sum
```
dexter2206, line 49: **Question**: Why do we write this explicitly? Isn't it as simple as `return count * unit_cost * multiplier`?

The same question for other sequences in this file.

tests/data/invalid_repetitions.yaml
```diff
@@ -0,0 +1,125 @@
...
+          multiplier: 1
+      resources:
+      - name: x
+        type: additive
+        value: 10
+      type: null
+    version: v1
+  description: Routine with repetition with resources.
+  problems:
+  - "Routine with repetition should not contain any resources: ('root',), resources: ['x']."
```
dexter2206, line 125: **nit**: No newline at the end of this file.

src/bartiq/symbolics/backend.py
```diff
@@ -45,14 +45,14 @@ def reserved_functions(self) -> Iterable[str]:
...
-    def define_function(self, expr: TExpr[T], /, func_name: str, function: Callable[[TExpr[T]], TExpr[T]]) -> TExpr[T]:
-        """Define an undefined function."""
+    def substitute(
+        self,
+        expr: TExpr[T],
+        /,
+        replacements: Mapping[str, TExpr[T]],
+        functions_map: Mapping[str, Callable[[TExpr[T]], TExpr[T]]] | None = None,
+    ) -> TExpr[T]:
+        """Substitute all occurrences of symbols and functions in expr with given replacements and functions_map."""
```
dexter2206, line 55: **Praise**: I love this change!

dexter2206 (changes requested):
Overall looks good, and I only left minor suggestions, with a following caveat:

After reviewing this and the QREF's counterpart of this PR I'm more and more convinced that it would be better if we had a separate class representing "loop" or a "routine with repetitions" (or at least if we had it in Bartiq). This is because a lot of the logic we perform for validation and dispatching (with `if`s) could be neatly hidden. For instance, we verify if routine with repetition has one child - but we could have an object that HAS to have one child (which would be validated by Pydantic). Anyway, we can continue with the current solution and later refactor it if we have some time to do so.

src/bartiq/_routine.py
```diff
@@ -24,6 +24,7 @@
 from qref.schema_v1 import PortV1, ResourceV1, RoutineV1
 from typing_extensions import Self, TypedDict
 
+from .repetitions import Repetition, _repetition_from_qref, _repetition_to_qref
```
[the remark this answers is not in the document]
mstechly: Done

src/bartiq/repetitions.py
```diff
@@ -0,0 +1,315 @@
...
+    def get_sum(self, expr: TExpr[T], count: TExpr[T], backend: SymbolicBackend[T]) -> TExpr[T]:
+        return 0.5 * count * (2 * self.initial_term + (count - 1) * self.difference) * expr
+
+    def get_prod(self, expr: TExpr[T], count: TExpr[T], backend: SymbolicBackend[T]) -> TExpr[T]:
+        return (
+            self.difference**count
+            * backend.as_expression(f"gamma({backend.serialize(self.initial_term / self.difference + count)})")
+            / backend.as_expression(f"gamma({backend.serialize(self.initial_term / self.difference)})")
+            * expr**count
+        )
```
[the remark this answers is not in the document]
mstechly: Yup, this works!

src/bartiq/repetitions.py
```diff
@@ -0,0 +1,315 @@
...
+            new_prod = backend.substitute(self.prod, inputs, functions_map)
+        new_num_terms = backend.substitute(self.num_terms_symbol, inputs, functions_map)
+
+        return replace(self, sum=new_sum, prod=new_prod, num_terms_symbol=new_num_terms)
+
+
+@dataclass(frozen=True)
+class CustomSequence(Generic[T]):
+    """Custom sequence.
+    For sequences which do not fall into categories defined in other classes, one can use a custom representation.
```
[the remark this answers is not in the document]
mstechly: Done

src/bartiq/repetitions.py
```diff
@@ -0,0 +1,315 @@
...
+    def substitute_symbols(
+        self, inputs: dict[str, TExpr[T]], backend: SymbolicBackend[T], functions_map=None
+    ) -> CustomSequence[T]:
+        if functions_map is None:
+            functions_map = {}
+
+        if self.iterator_symbol in inputs.values():
+            raise BartiqCompilationError(
+                f"Tried to replace symbol that's used as iterator symbol in a sequence: {self.iterator_symbol}."
+            )
```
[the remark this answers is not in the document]
mstechly: Actually, I think it shouldn't be in neither keys or values?
Cause you don't want to accidentally replace iterator with sth else, but you also don't want any other symbol to be replaced into the operator, right?
Will make sure tests cover both.

tests/compilation/test_repetitions.py
mstechly: I just wanted this to be dead simple and explicit.

tests/data/invalid_repetitions.yaml
mstechly: Done

dexter2206 (approved):

PR #147, opened 2024-11-28
fix: Substitution of functions is no longer sensitive to order of functions dictionary

### LINKED ISSUE #146: Nested functions may fail to evaluate, depending on the order of functions in `custom_funcs` dict
## Describe the bug
Consider an expression like "f(g(x))". Substituting, in one go for `x`, `f` and `g` should result in a numerical value (as long as `f` and `g` produce numerical outputs for numerical outputs). However, this is not the case if the following two conditions are satisfied at the same time:

- In the input to `backend.substitute` method, `f` precedes `g` **and**
- Function assigned to `f` fails for nonnumerical inputs, in which case it is replaced by a dummy function of the same name.

## How to Reproduce
The following snippet reproduces the issue. Observe that `f` will fail for symbolic inputs.

```python
from bartiq.symbolics import sympy_backend

f = lambda x: int(x) + 1 # int(x) triggers a failure for symbolic expressions, which is needed here
g = lambda x: x ** 2

expr = sympy_backend.as_expression("f(g(x))")

result_1 = sympy_backend.substitute(expr, {"x": 3}, {"f": f, "g": g})
result_2 = sympy_backend.substitute(expr, {"x": 3}, {"g": g, "f": f})

print(result_1)  # Prints f(9), f failed to evaluate
print(result_2)  # Prints 10, this is correct
```

## Expected behavior
Both prints statements above produce the same output. The result of `backend.substitute` should not depend on the order of substitutions in input.

## Actual behavior
The `f` fails to evaluate in the first example above.

## Environment:**
This bug is OS/environment independent.

dexter2206:
## Description

Closes #146 

The problem was solved by dynamically defining proper sympy functions that can be evaluated when needed instead of relying on a success of a single call to the user defined function.

Additionally, this PR contains a boy-scout cleanup of the warnings raised in all our tests. At least when tested locally, the tests don't emit any warnings. Most warnings were just filtered out (as they are known and expected warnings), two of them were fixed because they resulted from typo.

### DIFF
--- a/src/bartiq/symbolics/sympy_backends.py
+++ b/src/bartiq/symbolics/sympy_backends.py
@@ -95,6 +95,24 @@ def parse_to_sympy(expression: str, debug: bool = False) -> Expr:
     return parse(expression, interpreter=SympyInterpreter(debug=debug))
 
 
+def _sympify_function(func_name: str, func: Callable) -> type[sympy.Function]:
+    if not issubclass(type(func), sympy.Function):
+
+        def _eval_wrapper(cls, *args, **kwargs):
+            try:
+                return func(*args, **kwargs)
+            # The except claus here is intentionally broad, you never know what
+            # func can raise.
+            except Exception:
+                return None
+
+        sympy_func = type(func_name, (sympy.Function,), {"eval": classmethod(_eval_wrapper)})
+    else:
+        sympy_func = func
+
+    return sympy_func
+
+
 class SympyBackend:
 
     def __init__(self, parse_function: Callable[[str], Expr] = parse_to_sympy):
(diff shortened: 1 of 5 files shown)


src/bartiq/symbolics/sympy_backends.py
```diff
@@ -95,6 +95,24 @@ def parse_to_sympy(expression: str, debug: bool = False) -> Expr:
     return parse(expression, interpreter=SympyInterpreter(debug=debug))
 
 
+def _sympify_function(func_name: str, func: Callable) -> type[sympy.Function]:
+    if not isinstance(func, sympy.Function):
+
+        def _eval_wrapper(cls, *args, **kwargs):
+            try:
+                return func(*args, **kwargs)
+            # The except claus here is intentionally broad, you never know what
```
mstechly, line 104:
```suggestion
            # The except clause here is intentionally broad, you never know what
```

mstechly (approved):

PR #158, opened 2025-01-24
feat: feat: add qubit highwater calculations

mstechly:
## Description

Adds post-processing stages and option to calculate qubit highwater.

## Please verify that you have completed the following steps

- [X] I have self-reviewed my code.
- [X] I have included test cases validating introduced feature/fix.
- [X] I have updated documentation.

PR #165, opened 2025-02-01
feat: Implement qubit highwater computation

dexter2206:
## Description

Implements computation of highwater. Also includes a ton of typing improvements.

Builds on top of @mstechly 's PR (#158 ), retaining the original history of his commits.

### DIFF
--- a/src/bartiq/compilation/postprocessing.py
+++ b/src/bartiq/compilation/postprocessing.py
@@ -126,7 +126,7 @@ def _get_highwater_for_non_leaf(
     for child in TopologicalSorter(graph).static_order():
         ancestors = [node for node in tree if any(predecessor in node for predecessor in graph[child])]
 
-        # Starting nodes are connected to source node represented by None
+        # Starting nodes are connected to source node represented by (None,)
         if not ancestors:
             tree[(child,)] = (None,)
         # Children with one incoming connections are simply connected to their ancestor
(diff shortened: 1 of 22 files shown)


src/bartiq/compilation/postprocessing.py
```diff
@@ -42,6 +44,227 @@ def aggregate_resources(
...
+
+def add_qubit_highwater(
+    routine: CompiledRoutine[T],
+    backend: SymbolicBackend[T],
+    resource_name: str = "qubit_highwater",
+    ancillae_name: str = "local_ancillae",
+) -> CompiledRoutine[T]:
+    """Add information about qubit highwater to the routine.
+
+    Qubit highwater is the number of qubits needed for a particular subroutine, at the place where it's "widest".
```
mstechly, line 235: Maybe it's worth mentioning here, that this calculation is an upper bound, as there might be cases, where e.g.: some subroutines need ancilla qubits only for a shot period of time and the same qubits could be reused, but this is not something our model can handle.
?

src/bartiq/compilation/postprocessing.py
```diff
@@ -42,6 +44,227 @@ def aggregate_resources(
...
+
+    Args:
+        routine: The routine to which the resources will be added.
+        backend : Backend instance to use for handling expressions. Defaults to `sympy_backend`.
+        resource_name: name for the added resource. Defaults to `qubit_highwater`.
```
mstechly, line 240: Actually, I think it should be `"qubit_highwater"`, no? Same for ancillae.

src/bartiq/compilation/preprocessing.py
```diff
@@ -88,8 +84,9 @@ def propagate_child_resources(routine: Routine[T], backend: SymbolicBackend[T])
         res_name: Resource(
             name=res_name,
             type=ResourceType.multiplicative,
-            value=math.prod(
-                (backend.as_expression(f"{child_name}.{res_name}") for child_name in children),  # type: ignore
+            value=reduce(
```
mstechly, line 87: Neat!

mstechly (approved):

Issue #132, opened 2024-10-24
Remove legacy parsers


mstechly:
We still have a legacy parser that we had before AST parser has been implemented (https://github.com/PsiQ/bartiq/pull/87). 
We no longer need it and we should remove it.

The legacy parser is `parse_to_sympy` method in `symbolics/sympy_interpreter.py`.

PR #166, opened 2025-02-04
refactor: Issue 132 remove legacy parsers

Brendan-Reid1991:
## Description

PR for [issue 132: remove legacy parsers](https://github.com/PsiQ/bartiq/issues/132)

Straightforward fix, with minor changes to tests. Removed `parse_to_sympy` from `bartiq/symbolics/sympy_interpreter.py`, and wherever it was imported as `legacy_parse_to_sympy`. Removed one test from `test_sympy_interpreter.py`, and modified another to not test for the legacy parser. 

Bulk of the modifications were automatic black formatting changes. If there are issues with the formatting that has been done, I can revert it!

Tests pass locally, with coverage at 93%.

- [x] I have updated documentation. (AFAIK :  N/A)


Brendan-Reid1991:
@mstechly thank you for the feedback! digging a bit more through the code now, I am tempted to completely remove the `Interpreter` base class, unless it is likely to be used for something other than sympy. I am not doing that because that we could end up just reintroducing it later on down the line, but you have more of an idea of how the code will change going forward. 

NB: previous push was down erroneously, tests are failing still!

dexter2206:
> @mstechly thank you for the feedback! digging a bit more through the code now, I am tempted to completely remove the `Interpreter` base class, unless it is likely to be used for something other than sympy. I am not doing that because that we could end up just reintroducing it later on down the line, but you have more of an idea of how the code will change going forward.
> 
> NB: previous push was down erroneously, tests are failing still!

The `Interpreter` base class defines the interface used by parsers. In particular, instances of Interpreter are what `ast_parser.parse` expects as input. This interface is important if we want to support other types of expressions - and we certainly have to at some point, given that we want to implement nested expression trees. So, I strongly recommend that we keep this class.

That being said, after legacy parser is removed, the interface of `Interface` could be simplified (it was designed for legacy parser, and then used in `ast_parser`). I'll create a separate issue for this.

src/bartiq/symbolics/ast_parser.py
```diff
@@ -129,23 +129,23 @@ def _replace_wildcards(expression):
 _WILDCARD_REPLACEMENT = _PreprocessingStage(matches=_contains_wildcard, preprocess=_replace_wildcards)
 
 
-def _contains_lambda(expression):
+def _contains_lambda(expression: str) -> bool:
```
mstechly, line 132: [comment] FYI, we don't require typing for private functions, but it's obviously always good to have it added, thanks!

src/bartiq/symbolics/grammar.py
```diff
@@ -15,96 +15,13 @@
...
-        (term + ZeroOrMore(Group(add_sub_op + term)))
-        .set_name("expression")
-        .set_parse_action(interpreter.create_expression)
-    )
-
-    # Define complete expression
-    complete_expression = (Suppress(StringStart()) + expression + Suppress(StringEnd())).set_name("complete expression")
-
-    return complete_expression
+from pyparsing import ParseResults
```
mstechly, line 18: [question] Do we still need it?
It's used in `debuggable`, not sure if we even trigger this condition now that all the other stuff from pyparsing is gone?
If not, it seems we could get rid of `pyparsing` dependency alltogether?

tests/symbolics/test_grammar.py
```diff
@@ -61,11 +52,3 @@ def create_number(self, tokens):
     @debuggable
     def create_function(self, tokens):
         """Dummy method."""
-
-    @debuggable
```
mstechly, line 65: [question] Was this simply not used anywhere?

mstechly (approved):
Some minor questions, but otherwise looks good to me!

src/bartiq/symbolics/grammar.py
```diff
@@ -15,96 +15,13 @@
...
 
 
 class Interpreter(ABC):
```
mstechly, line 21: [suggestion] Given that this file now mostly contains `Interpreter` class, perhaps it's worth renaming it to `interpreter.py` ?

tests/symbolics/test_grammar.py
Brendan-Reid1991: Yeah, since create_expression and create_unary_atom seemed to only be used within the legacy parser, I removed them. Conceivably they might need to be reintroduced later, but it seemed to make sense to remove them if there is no immediate use case, as they were just abstract methods.

src/bartiq/symbolics/grammar.py
Brendan-Reid1991: done!

src/bartiq/symbolics/grammar.py
Brendan-Reid1991: I have also removed the pyparsing installation from pyproject, but I understand if you want to do that change under another PR!

Issue #26, opened 2024-05-03
Remove duplicate function names


mstechly:
We discussed with @dexter2206 that we should remove any duplicate keywords from `sympy_interpreter`. Currently we have many ways to write same functions, e.g.: `log2` and `log_2`, `ceil` and `ceiling`. 
It would be better to have only one way as it will reduce confusion and set expectation more clearly.

PR #167, opened 2025-02-12
refactor: remove duplicate function names

Brendan-Reid1991:
## Description

PR for [Issue 26](https://github.com/PsiQ/bartiq/issues/26). 

 Removed `ceil` in favour of `ceiling` and `log_x` in favour of `logx`. 

Following the comments from @dexter2206  and @mstechly, I updated this PR with some significant changes. Taking Konrads idea I changed the behaviour of the function `validate_expression` to `find_undefined_functions` which returns a list of tuples of `(potentially unknown functions, suggested replacement)`. I also added this into the `SymbolicBackend` protocol. 

I think this is a change for the better, however what is now missing from this PR is any kind of hinting or help to the user about how to use this functionality. At present, this requires the user to simply *know* a) what the backend is and b) that this function exists. I think this could feasibly exist as part of a broader debugging suite, but that sounds like a bigger question and outside of the scope of this PR. 


src/bartiq/symbolics/sympy_backends.py
```diff
@@ -275,6 +279,58 @@ def sequence_prod(
         """Express a product of terms expressed using `iterator_symbol`."""
         return sympy.Product(term, (iterator_symbol, start, end))
 
+    @staticmethod
+    def inspect_expression(expression: sympy.Basic) -> None:
```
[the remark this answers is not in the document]
mstechly: Also, the name is somewhat confusing.
If things go well, it doesn't return anything nor it raises anything. 
Perhaps `check_expression` or `validate_expression` could be better?

src/bartiq/symbolics/sympy_backends.py
```diff
@@ -275,6 +279,58 @@ def sequence_prod(
...
+        """
+        ops = _unpack_expression_into_operations(expression=expression)
+
+        known_functions: set[str] = set(_ALL_SYMPY_CONSTANTS + _ALL_SYMPY_OPS + BUILT_IN_FUNCTIONS)
+        potentially_unknown_functions: list[str] = list(ops - known_functions)
+        if potentially_unknown_functions:
+            closest_match: list[str] = difflib.get_close_matches(
+                word=potentially_unknown_functions[0], possibilities=known_functions
+            )
+            msg = f"Unrecognised function call '{potentially_unknown_functions[0]}'."
```
mstechly, line 304: It only acts on the first function.
Consider the following expression:
`g(7) + log_2(10) + 5`.
It will fail on `g(7)`, even though the user might actually not care about it and doesn't want to evaluate.
But it will fail to notify user about the log_2. 
We could return an error message that will just list all the functions? Or provide an option to add "user defined known functions", which are not built in?

src/bartiq/symbolics/sympy_backends.py
```diff
@@ -275,6 +279,58 @@ def sequence_prod(
...
+            if closest_match:
+                msg += f"\nDid you mean {f"one of {closest_match}" if len(
+                    closest_match) > 1 else f"'{closest_match[0]}'"}?."
+
+            raise ValueError(msg)
```
mstechly, line 309: [question] Since we don't use `self` anywhere here, I don't see why it should be part of the `SymbolicBackend` class and not a standalone function in the `sympy_backends` module?

src/bartiq/symbolics/sympy_backends.py
mstechly, line 1: Why it's called `sympy_backends` and not `sympy_backend`?
@dexter2206?

mstechly (changes requested):

src/bartiq/symbolics/sympy_backends.py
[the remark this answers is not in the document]
Brendan-Reid1991: Yes, I didn't write tests in case we decided to scrap it entirely! And In terms of returning nothing, one thought I had was to `print` a message saying something along the lines of "No problems found in the given expression. If you think this is a bug, please open an issue on the GitHub" with a link. 

Very open to changing the name too, I agree it is not informative.

src/bartiq/symbolics/sympy_backends.py
Brendan-Reid1991: aha, this is an excellent suggestion. Truthfully I hadn't considered user-defined functions, so only taking the first arg of `potentiall_unknown_functions` seemed reasonable, but allowing the user to provide a list of function names to ignore, I love this idea.

src/bartiq/symbolics/sympy_backends.py
Brendan-Reid1991: My thinking was that if the user needed something like this, they most likely had imported the `sympy_backend` object already to build their routine, or they are at least aware that such an object exists in order for their routine to compile. In an interactive environment (which is where I see this method getting the most use) this means their kernel already has the tool imported, rather than having to run another command to import the function from the package. 

The difference between the two is pretty minor, I think, because the user will still have to know such a method / function exists in order to use it.

src/bartiq/symbolics/sympy_backends.py
```diff
@@ -275,6 +279,58 @@ def sequence_prod(
...
+
+
+def _unpack_expression_into_operations(expression: sympy.Basic) -> set[str]:
```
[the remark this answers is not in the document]
Brendan-Reid1991: In other locations `sympy.Basic` makes sense because I need to catch `Expr` and `Function` and `Zero`, but here you're right that here, this could just be `Expr`.

src/bartiq/symbolics/sympy_backends.py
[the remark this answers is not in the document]
mstechly: I think having a print would be a good idea 👍

src/bartiq/symbolics/sympy_backends.py
mstechly: Ok, sounds good 👌

src/bartiq/symbolics/sympy_backends.py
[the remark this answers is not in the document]
Brendan-Reid1991: Sorry, just seeing exactly where this is! in `_unpack_expression_into_operations` it does need to be `Basic` for the reasons I mentioned, the expression tree in sympy can have all manner of objects in it, but for the main function (now called `validate_expression`) it only needs to be a `Expr`, so it has been changed up there.

src/bartiq/symbolics/sympy_backends.py
dexter2206: Because we had two backends. We can rename it now.

src/bartiq/symbolics/sympy_backends.py
dexter2206: Actually most users don't have `sympy_backend` imported. The reason is that usually the default argument for `backend` is used in `compile_routine` and `evaluate`, i.e. most people will do:
```python
result = compile(routine) # no mention of backend here
```
instead of
```python
result = compile(routine, backend=sympy_backend)
```

src/bartiq/symbolics/sympy_backends.py
```diff
@@ -18,9 +18,10 @@
 
 from __future__ import annotations
 
+import difflib
```
dexter2206, line 21: **praise**: Using `difflib` to find matches is a very neat idea!

src/bartiq/symbolics/sympy_interpreter.py
```diff
@@ -185,17 +185,14 @@ def eval(cls, n):
     "prod": better_prod,
     "exp": exp,
     "log": log,
-    "ceil": ceiling,
```
mstechly, line 188: I'm not sure about this one.
On one hand, `ceiling` is used in `sympy`, on the other had both `math` and `numpy` have `ceil` methods, so it feels a bit more "pythonic"?
I'm fine either way, but I'm not sure which way is better ;)
@dexter2206 thoughts?

Brendan-Reid1991: Agreed, this one was a bit of a toss up for me. Happy with either, Konrad gets the deciding vote :p

src/bartiq/symbolics/sympy_backends.py
mstechly: @Brendan-Reid1991 could you rename this?

src/bartiq/symbolics/sympy_interpreter.py
mstechly: @dexter2206 WDYT?

mstechly (approved):

src/bartiq/symbolics/sympy_interpreter.py
dexter2206: Let's go with `ceil` to match the `ceil` from standard lib.

src/bartiq/symbolics/backend.py
```diff
@@ -140,3 +140,10 @@ def sequence_sum(self, term: TExpr[T], iterator_symbol: TExpr[T], start: TExpr[T
 
     def sequence_prod(self, term: TExpr[T], iterator_symbol: TExpr[T], start: TExpr[T], end: TExpr[T]) -> TExpr[T]:
         """Express a product of terms expressed using `iterator_symbol`."""
+
+    def find_undefined_functions(
+        self, term: TExpr[T], user_defined: Optional[list[str]] = None
```
dexter2206, line 145: **style**: There are two things related to type hints that can be improved here:
- The `x | None` syntax is preferred over `Optional[x]` for Python 3.10+ ([source](https://github.com/python/cpython/pull/30222))
- Using `list` for `user_defined` is unnecessary narrow. We are not planning on `append`ing to `user_defined`, or doing anything specific to a list. I think that a reasonable type would be `Iterable[str]` (with `Iterable` from `collections.abc`). Similarly, let's reconsider if we want to require that the return type is necessarily a list, maybe an iterable is sufficient as well?
- Combining the previous points, it would seem `Optional` is not needed at all, and we could just use something like `user_defined: Iterable[str] = ()`.

src/bartiq/symbolics/sympy_backends.py
```diff
@@ -275,6 +276,52 @@ def sequence_prod(
         """Express a product of terms expressed using `iterator_symbol`."""
         return sympy.Product(term, (iterator_symbol, start, end))
 
+    def find_undefined_functions(
+        self, expr: sympy.Expr, user_defined: Optional[list[str]] = None
```
dexter2206, line 280: **style**: If we decided to improve typing in `backend.py` we also have to make matching change here.

Also, I think we have a subtle typing error here, that surprisingly doesn't trigger mypy errors:
- The type of `expr` in the backend interface is `TExpr[T]`
- The `T` for `SympyBackend` is `sympy.Expr`
- Therefore, the `expr` should be annotated as `TExpr[Expr]` to match the interface OR the interface should be changed so that `expr` has type `T`. My hunch is that the first option is better here, otherwise it could cause some friction if the user wanted to use expressions produced in compilation without manually casting them to `T`.

Also 2: The docstrings should not contain types, as they are already present in the signature.

dexter2206 (changes requested):
I think this is good to go as soon as we fix (very minor) typing issues!

dexter2206 (approved):
LGTM

PR #169, opened 2025-02-18
Make highwater sensitive to chronology

dexter2206:
## Description

The current implementation of highwater assumes some degree of parallelism, which conflicts with how we compute the highwater in other tools. This PR switches to an implementation which takes into account the chronology of routines.

To facilitate the change, I added `children_order` field to classes representing routines. This is because relying on the order of dictionaries is prone to errors. The field stores a tuple of children names marking their ordering. When constructing a routine from  QREF, the order is the same as in the QREF file.

The QREF itself does not impose restrictions on how the children are sorted. Therefore, we only assume that the `child_order`  represents chronology iff it also matches one of possible topological orderings. If this is not the case, we take one of topological orderings as the one representing the chronology, and we warn the user about the possible inaccuracies.

The PR also includes some boy scout changes improving typing.


pyproject.toml
```diff
@@ -11,7 +11,7 @@ readme = "README.md"
 python = "^3.10"
 pydantic = "^2.7"
 sympy = "^1.12"
-qref = "0.9.0"
+qref = "0.10.0"
```
mstechly, line 14: Why not `^0.10.0`?

src/bartiq/compilation/postprocessing.py
```diff
@@ -49,181 +47,36 @@ def _inner(routine: CompiledRoutine[T], backend: SymbolicBackend[T]) -> Compiled
...
-
-    # Include local ancillae as well.
+    watermarks.append(_outflow(routine))
     local_ancillae = routine.resources[ancillae_name].value if ancillae_name in routine.resources else 0
 
-    # And finally combine the results.
-    children_highwater = backend.max(*[highwater_map[key] for key in tree])
-    return children_highwater + local_ancillae + passthrough_cost
+    nonzero_watermarks = [watermark for watermark in watermarks if watermark != 0]
+    return backend.max(*nonzero_watermarks) + local_ancillae
```
mstechly, line 76: Wouldn't it behave exactly the same way if you used `watermarks` instead of `non_zero_watermarks`?

mstechly (approved):

src/bartiq/compilation/postprocessing.py
dexter2206: Excellent question. No it wouldn't, because currently we don't have assumptions about the sign of parameters. Suppose `watermarks == [0, Symbol("N")]`. We know that the symbol `N` here has to be positive (or at the very least nonnegative), but sympy doesn't, and so `backend.max(*watermarks)` would not exclude the zero.

Obviously this is only relevant when examining symbolic expressions, after the evaluation the result should be the same. I just didn't want to contribute to "gnarliness" of our expressions.

pyproject.toml
dexter2206: I followed the convention we used previously. However, I agree that caret specifier would be better, so I changed it in https://github.com/PsiQ/bartiq/pull/169/commits/d675694c323a574aebe191782338a12fa8208c92

PR #175, opened 2025-03-04
Fixing log2 issue

Brendan-Reid1991:
## Description

Fixing the "log2 issue" where logarithms of X in base-2 are reported as natural logs of X divided by log(2). 

This PR modifies the `substitute` method in `sympy_backend` with a `_postprocess` call if the expression cannot be forced into a numeric form. 

`_postprocess` applies a series of "post processing steps", functions that take in a `sympy.Expr` and output a `sympy.Expr`. At present, the only post processing step is this fix on logarithms, and _only if_ the expression has a `1/log(2)` subexpression, although more could be feasibly added.

`_correct_base2_logs` is the function doing the bulk of the work, which recursively unpacks the arguments of the given expression and checks if each matches a specific pattern (`_LOG2_EXPR`). If it does, the output from `.match` is passed into `_LOG2_REPL` in place of the original argument. Else, the argument itself is returned. Similarly if the argument is of type `sympy.Symbol` or `sympy.Number`.

The reason this implementation was chosen was because global changes or detecting patterns from the top level expression were too unreliable, and considering the complex expressions that can come out of Bartiq they were not feasible as solutions. The 'check' on an expression within `_postprocess`, where `_correct_base2_logs` is only added to the list `post_processing_steps` if the expression contains `1/log(2)` somewhere, is a small step towards avoiding needless evaluation. 

The downside is that there is no way (that I know of) to prevent the recursive function from going "too deep" in the expression tree, i.e. it will fully unpack and check each argument and sub argument of
```
ceiling(P*(1 + x*y + z) / log(2))
```
solely because it _has_ a 1/log(2) expression. Even using wildcards, I could not find a way to reliably get the `has` method to work with something like `Wild(X)/log(2)` or `log(Wild(X))/log(2)`, i.e.:

```
from sympy import sympify, log, Wild
expr = sympify("ceiling(log(x)/log(2))")
print(expr.has(1/log(2)))
>>> True
print(expr.has(Wild('X')/log(2)))
>>> False
print(expr.has(log(Wild('X'))/log(2)))
>>> False
```

## Performance Penalities
Running the full pytest suite via `pytest` in the terminal on a Mac M4
- Previously: 28.6s
- With this PR: 34.9s
Increase of ~22%

Running `cProfile` on the full pytest suite
- Previously: 88s
- With this PR: 104.5s
~ 18% increase.

- [x]  I have included test cases validating introduced feature/fix.
- [ ] I have updated documentation. (N/A I believe)


tests/symbolics/test_sympy_backend.py
```diff
@@ -178,3 +183,54 @@ def test_min_max_works_for_symbols(backend):
     values = symbols("a, b, c")
     assert backend.min(*values) == sympy_min(*values)
     assert backend.max(*values) == sympy_max(*values)
+
+
+def test_log2_expression():
```
mstechly, line 188: [issue] Why are we testing private functions here?
In general private functions should not be tested, because they should be an implementation detail. So if we decide to change them in future, we'll be forced to change tests as well. Tests should only. check the behaviour that's observable by the user and private functions by definition aren't really observable.

mstechly (changes requested):
My main issue with how it's being tested is that we actually don't have any tests for observable behavior (see my other comment).

PR #176, opened 2025-03-21
feat: adds derived resources and moves highwater from postprocessing

mstechly:
## Description

Adds `derived_resources` to compilation and moves calculating `qubit_highwater` from postprocessing to `derived_resources`. 

- [X] I have included test cases validating introduced feature/fix.
- [X] I have updated documentation.


src/bartiq/compilation/_compile.py
```diff
@@ -107,7 +131,8 @@ def compile_routine(
         preprocessing_stages: functions used for preprocessing of a given routine to make sure it can be correctly
             compiled by Bartiq.
         postprocessing_stages: functions used for postprocessing of a given routine after compilation is done.
-        skip_verification: a flag indicating whether verification of the routine should be skipped.
+        derived_resources: collection containing information needed to calculate derived resources.
```
dexter2206, line 134: **nitpick**: It would be great if we could give more information on the structure of this iterable. Something like "iterable with dictionaries describing how to calculate derived resources. Each dictionary should contain the derived resource's name, type and the function mapping a routine to the value of resource."

src/bartiq/compilation/_compile.py
```diff
@@ -316,3 +343,37 @@ def _compile(
         repetition=repetition,
         children_order=routine.children_order,
     )
+    return _add_derived_resources(compiled_routine, derived_resources, backend)
+
+
+def _accepts_resource_name(func: Calculate[T] | CalculateWithName[T]) -> TypeIs[CalculateWithName[T]]:
```
dexter2206, line 349: **praise**: Nice usage of `TypeIs`!

src/bartiq/compilation/_compile.py
```diff
@@ -94,6 +117,7 @@ def compile_routine(
     backend: SymbolicBackend[T] = sympy_backend,
     preprocessing_stages: Iterable[PreprocessingStage[T]] = DEFAULT_PREPROCESSING_STAGES,
     postprocessing_stages: Iterable[PostprocessingStage[T]] = DEFAULT_POSTPROCESSING_STAGES,
+    derived_resources: Iterable[DerivedResources] | None = None,
```
dexter2206, line 120: **typing/style**: Having `None` here works, but consider how we use `derived_resources`. We only iterate over it, and we don't really need to take special actions if it's empty, right? Therefore, we could simplify this type annotation while simultaneously getting rid of some conditional logic later on.

```suggestion
    derived_resources: Iterable[DerivedResources] = (),
```

src/bartiq/compilation/_compile.py
```diff
@@ -316,3 +343,37 @@ def _compile(
...
     )
+    return _add_derived_resources(compiled_routine, derived_resources, backend)
+
+
+def _accepts_resource_name(func: Calculate[T] | CalculateWithName[T]) -> TypeIs[CalculateWithName[T]]:
+    return "resource_name" in inspect.signature(func).parameters
+
+
+def _add_derived_resources(
+    routine: CompiledRoutine[T], derived_resources: Iterable[DerivedResources[T]] | None, backend: SymbolicBackend[T]
```
dexter2206, line 354: **style**: (continuation of previous comment) Same annotation works here.

```suggestion
    routine: CompiledRoutine[T], derived_resources: Iterable[DerivedResources[T]], backend: SymbolicBackend[T]
```

src/bartiq/compilation/_compile.py
```diff
@@ -316,3 +343,37 @@ def _compile(
...
+) -> CompiledRoutine[T]:
+    if derived_resources is None:
+        return routine
```
dexter2206, line 357: **style**: And if the previous comments are applied, this `if` statement can just be removed.

dexter2206 (changes requested):
Left some comments regarding style/typing, but otherwise LGTM!

src/bartiq/compilation/_compile.py
mstechly: Done!

src/bartiq/compilation/_compile.py
mstechly: Neat trick, fixed!

dexter2206 (approved):

Issue #138, opened 2024-11-07
refactor bartiq to use single symbol for expressing `pow` method


mstechly:
Currently in Bartiq we allow to express `pow` function as both `^` and `**`. 
This is problematic since:
- `^` in Python is a XOR operator, 
- an expression including one symbol could be later serialized and the symbol might turn into a different one, which could be confusing.

After [some discussion with @dexter2206](https://github.com/PsiQ/qref/pull/124#discussion_r1830899552), we think that using `**` would be better than `^`.

PR #187, opened 2025-04-20
Refactoring pow operations

jyotiraj-code:
## Description

Fixing #138 

- Context: Discussion to refactor Bartiq to represent `pow` operations with `**` instead of `^` in [Here](https://github.com/PsiQ/qref/pull/124#discussion_r1830899552)
- definition `_contains_xor_op()`  and `_replace_xor_op()` was provided already to fix this issue

Fixed:
- `ast_parser.py` still contain `ast.BitXor: operator.pow` which was changed to `xor` operation
- Updated `sympy_interpreter.py` operator method for `^`

### DIFF
--- a/src/bartiq/symbolics/ast_parser.py
+++ b/src/bartiq/symbolics/ast_parser.py
@@ -60,7 +60,7 @@ _BINARY_OP_MAP = {
     ast.Sub: operator.sub,
     ast.Mod: operator.mod,
     ast.Pow: operator.pow,
-    ast.BitXor: operator.pow,
+    ast.BitXor: operator.xor,
     ast.FloorDiv: operator.floordiv,
 }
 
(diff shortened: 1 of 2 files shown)


mstechly:
It seems to me that we use a `^` sign in a couple of places in the tests (e.g. in `repetitions.yaml`). This is slightly surprising, as I would expect these tests to fail. 
So I'd suggest the following changes:
- Write a test that will check if a routine with a resource having `^` works as expected (i.e. does xor)
- Update test cases in `repetitions.yaml` to use `**` instead of `^`
- Investigate why these tests pass when they still have `^`?
- Also, I think it would be worth changing the `_print_Pow` method in `sympy_serializer` to use `**` instead of `^`.

mstechly:
Closing due to inactivity.

PR #194, opened 2025-05-07
Allow transitive compilation of resources

Brendan-Reid1991:
## Description

This PR adds a keyword to `bartiq._compile.compile`, `transitive_resources: bool`, which allows "partial compiation" of a routine. Rather than the root and it's children having their expressions defined fully in the backend expression type (and the simplifications that occur in the background), they are instead defined "transitively" such that a resource will have a value like `child_a.resource + child_b.resource`. During evaluation then, these transitive resources are unwrapped and full, complete expressions are created. 

For example, given an uncompiled routine `A`, the following is valid:
```
from bartiq import compile_routine, evaluate

transitive_compilation: CompiledRoutine = compile_routine(A, transitive_resources=True)
full_compilation: CompiledRoutine = compile_routine(A, transitive_resources=False)

transitive_compilation == full_compilation
>>> False

evaluate(transitive_compilation, {}).routine == full_compilation
>>> True

```


dexter2206 (approved):
Good stuff.

src/bartiq/compilation/_compile.py
```diff
@@ -135,6 +142,8 @@ def compile_routine(
             Each dictionary should contain the derived resource's name, type
             and the function mapping a routine to the value of resource.
         skip_verification: flag indicating whether verification of the routine should be skipped.
+        transitive_resources: flag indicating if resource expressions should be compiled only transitively, i.e.
```
mstechly, line 145: I don't like the name here.
`transitive_resources`, as it's a noun which sounds like an actual object you would pass and not just a flag.
I'm trying to think of a better name, but it's hard. `transitive_resources_flag` is a bit long + `skip_verification` doesn't have a word "flag".
Maybe `skip_resource_unrolling`?
Thoughts?

src/bartiq/compilation/_compile.py
```diff
@@ -223,11 +242,11 @@ def _process_repeated_resources(
         assert resource_name in child_resources
         assert backend.serialize(resource.value) == f"{children[0].name}.{resource.name}"
     for resource in child_resources.values():
-        if resource.type == "additive":
+        if resource.type == ResourceType.additive:
```
mstechly, line 245: Thank you!

tests/compilation/data/compile/transitive_resources.yaml
```diff
@@ -0,0 +1,1645 @@
+- - program:
```
mstechly, line 1: **suggestion**: It would be nice to have some short comments describing each test case.
It's not obvious how the first two test cases are different from each other from just looking at them.

tests/compilation/test_compile.py
```diff
@@ -13,36 +13,37 @@
...
-def test_compile(routine, expected_routine, backend):
-    compiled_routine = compile_routine(routine, skip_verification=False, backend=backend).to_qref()
+def test_compile_with_no_transitive_resources(routine, expected_routine, backend):
+    compiled_routine = compile_routine(
+        routine, skip_verification=False, backend=backend, transitive_resources=False
+    ).to_qref()
     assert compiled_routine == expected_routine
 
 
+@pytest.mark.filterwarnings("ignore:Found the following issues with the provided routine")
```
mstechly, line 38: **question**: Why do we need this ignore?

tests/compilation/test_compile.py
```diff
@@ -249,20 +272,13 @@ def test_compilation_works_as_expected_in_presence_of_large_number_of_children(b
...
     assert (
         list(compilation_result.resources) == ["t_count", "foo"]
-        and backend.serialize(compilation_result.resources["t_count"].value) == expected_t_count
-        and backend.serialize(compilation_result.resources["foo"].value) == expected_foo
+        and backend.serialize(compilation_result.resources["t_count"].value) == backend.serialize(expected_t_count)
+        and backend.serialize(compilation_result.resources["foo"].value) == backend.serialize(expected_foo)
     )
+
+
+if __name__ == "__main__":
```
mstechly, line 283: Not needed, please remove.

tests/compilation/test_compile.py
```diff
@@ -249,20 +272,13 @@ def test_compilation_works_as_expected_in_presence_of_large_number_of_children(b
...
-    # expected_t_count = " + ".join(f"{child_name}.t_count" for child_name in sorted_child_names)
-    # expected_foo = "*".join(f"{child_name}.foo" for child_name in sorted_child_names)
-    compilation_result = compile_routine(routine, backend=backend).routine
-
+    compilation_result = compile_routine(routine, backend=backend, transitive_resources=transitive_resources).routine
     assert (
         list(compilation_result.resources) == ["t_count", "foo"]
-        and backend.serialize(compilation_result.resources["t_count"].value) == expected_t_count
-        and backend.serialize(compilation_result.resources["foo"].value) == expected_foo
+        and backend.serialize(compilation_result.resources["t_count"].value) == backend.serialize(expected_t_count)
```
mstechly, line 278: **question**: Why do you need to serialize it if it's a string?

tests/compilation/test_derived_resources.py
```diff
@@ -119,7 +119,9 @@ def test_highwater_of_an_empty_routine_is_zero():
     assert compiled_routine.resources["qubit_highwater"].value == 0
 
 
-def test_additive_derived_resources_are_processed_correctly():
+@pytest.mark.parametrize("transitive_resources", [True, False])
+def test_additive_derived_resources_are_processed_correctly(transitive_resources):
+    test_resource = "test_resource"
```
mstechly, line 124: **nitpick**: I'd rename it to `resource_name`.

tests/compilation/test_derived_resources.py
```diff
@@ -138,20 +140,27 @@ def test_additive_derived_resources_are_processed_correctly():
...
+    derived_resources = [{"name": test_resource, "type": "additive", "calculate": add_test_resource}]
+    compilation_result = compile_routine(
+        input_routine, derived_resources=derived_resources, transitive_resources=transitive_resources
+    )
+    compiled_routine = compilation_result.routine
+    backend = compilation_result._backend
+    assert (
+        compiled_routine.resources[test_resource].value == 17
+        if not transitive_resources
+        else backend.as_expression("+".join(f"{child}.{test_resource}" for child in compiled_routine.children))
```
mstechly, line 165: **question**: For non-transitive case, I'd also expect the top-level to evaluate to a number, no?

tests/compilation/test_repetitions.py
```diff
@@ -253,18 +253,23 @@ def test_closed_form_sequence_works_when_sum_and_prod_unspecified(sum_none, prod
             compiled_routine = compile_routine(routine).routine
 
 
-def test_custom_sequence_throws_error_when_replacing_iterator_symbol(backend):
+@pytest.mark.parametrize("transitive_resources", [True, False])
```
mstechly, line 256: **issue**: I don't see the point of adding `transitive_resources=True` to this test case, if we have to add if statements to handle it.
It's not expected behaviour that we throw an error here, so we don't need to test it 🤷‍♂️

mstechly (changes requested):

src/bartiq/compilation/_compile.py
Brendan-Reid1991: How about `allow_transitive_resources`?

tests/compilation/test_compile.py
Brendan-Reid1991: It was included in the above test case `test_compile_with_no_transitive_resources` (previously just `test_compile`) which I copied. I took it for granted that it was there for a good reason! But it seems that removing that warning from both tests does not cause any issue.

tests/compilation/test_compile.py
Brendan-Reid1991: Yeah there is no need to, this is a mistake 😅

tests/compilation/test_derived_resources.py
Brendan-Reid1991: For the *non-transitive* case it does evaluate to a number, but for the transitive case we should not expect it to evaluate to a number.

tests/compilation/test_repetitions.py
Brendan-Reid1991: This was one of the tests that failed when I implemented the changes, so I figured it should pass for both the previous behaviour and the new, default behaviour. This test is two-in-one (as it is checking for two `raises`) and the second error is raised correctly, so I'm happy to split this into two tests where one is only for non-transitive resources.

tests/compilation/data/compile/transitive_resources.yaml
Brendan-Reid1991: done!

src/bartiq/compilation/_compile.py
Brendan-Reid1991: Alternatives:
- `transitive_compile =True`
- `compact_resource_expressions = True`
- `transitive_expressions = True`
- `resource_expressions: Literal['transitive', 'full'] = 'transitive'`

mstechly: I think `allow_transitive_resources` sounds good. @dexter2206 ?

tests/compilation/test_derived_resources.py
mstechly: Oh ok, I got confused, looks good 👌

mstechly (approved):

PR #197, opened 2025-05-08
feat: created `CompilationFlags` enum

Brendan-Reid1991:
## Description

This PR adds `CompilationFlags`, a `enum.Flag` class that are used to modify the compilation process. 

This replaces the use of the keywords `skip_verification` and `allow_transitive_resources` in `bartiq.compilation._compile.compile`. 
- Instead of `skip_verification=True`, use `compilation_flags = CompilationFlags.SKIP_VERIFICATION`
- Instead of `allow_transitive_resources=False`, use `compilation_flags = CompilationFlags.EXPAND_RESOURCES`.
Flags can be combined bitwise, i.e. `compilation_flags = FLAG1 | FLAG2`.

Other minor updates:
- Fixed a bug where transitive resources were not being evaluated correctly in repetitions. 
- Added `bartiq.compilation` to the docs API reference.

### DIFF
--- a/src/bartiq/compilation/_compile.py
+++ b/src/bartiq/compilation/_compile.py
@@ -294,7 +294,7 @@ def _compile(
     inputs: dict[str, TExpr[T]],
     context: Context,
     derived_resources: Iterable[DerivedResources] = (),
-    compilation_flags: CompilationFlags = CompilationFlags(0),
+    compilation_flags: CompilationFlags = CompilationFlags(0),  # CompilationsFlags(0) corresponds to no flags
 ) -> CompiledRoutine[T]:
     try:
         new_constraints = evaluate_constraints(routine.constraints, inputs, backend)
--- a/src/bartiq/compilation/_compile.py
+++ b/src/bartiq/compilation/_compile.py
@@ -20,6 +20,7 @@ import warnings
 from collections import defaultdict
 from collections.abc import Iterable, Sequence
 from dataclasses import dataclass, replace
+from enum import Flag, auto
 from graphlib import TopologicalSorter
 from typing import Generic, Protocol
 
@@ -74,8 +75,15 @@ REPETITION_ALLOW_ARBITRARY_RESOURCES_ENV = "BARTIQ_REPETITION_ALLOW_ARBITRARY_RE
 # For instance, the following parameter tree:
 # {None: {"#out_0": N}}
 # tells us that the output port of the routine currently being
-# procesed should have size set to N.
+# processed should have size set to N.
 ParameterTree = dict[str | None, dict[str, TExpr[T]]]
+
+
+class CompilationFlags(Flag):
+    """A collection of compilation flags to modify `compile_routine` functionality."""
+
+    EXPAND_RESOURCES = auto()
+    """Expand resource values into full, rather than transitive, expressions."""
 
 
 class Calculate(Protocol[T]):
@@ -125,7 +133,7 @@ def compile_routine(
     postprocessing_stages: Iterable[PostprocessingStage[T]] = DEFAULT_POSTPROCESSING_STAGES,
     derived_resources: Iterable[DerivedResources] = (),
     skip_verification: bool = False,
-    allow_transitive_resources: bool = True,
+    compilation_flags: CompilationFlags | None = None,
 ) -> CompilationResult[T]:
     """Performs symbolic compilation of a given routine.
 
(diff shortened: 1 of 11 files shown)


mstechly:
I wasn't sure at first, but it actually is quite neat!
One thing that's worth considering is being able to set these compilation flags as strings. Could we do that with multiple at once?
Cause I think that since we want to keep the API simple, I'd prefer not to force people to import `CompilationFlags`.
Thoughts?

dexter2206:
> One thing that's worth considering is being able to set these compilation flags as strings. 

Please don't. From my experience solutions like this often evolve into "stringly typed systems", where slowly everything can become a string.

A concrete argument against using strings: this actually discards some advantages of static  analysis. Consider a user who mistakenly uses "UNKNOWN_FLAG". Then this fails before the code runs, and can be even highlighted by the language server as an error:

```python
compile_routine(routine, flags=CompilationFlags.UNKNOWN_FLAG | CompilationnFlags.LEGIT_FLAG)
```
The message here would be clear: there is no such thing as `CompilationFlags.UNKNOWN_FLAG`. This also fails at runtime, with similarly clear `AttributeError`.

On the other hand, this fails only at runtime, and only if we get the implementation right:

```python
compile_routine(routine, flags="UNKNOWN_FLAG | LEGIT_FLAG")
```

Another point from the tooling perspective is lack of autocomplete. `CompilationFlags.` can be completed to see the available options. With strings you have to either memorize or open the docs. Expect people constantly confused if it was "EXPAND_RESOURCES", "EXPANDED_RESOURCES" etc.

Flags are flags, strings are strings, let's keep them separated.

mstechly:
Ok, I'm convinced :D

src/bartiq/compilation/_compile.py
```diff
@@ -74,10 +75,20 @@
...
 # tells us that the output port of the routine currently being
-# procesed should have size set to N.
+# processed should have size set to N.
 ParameterTree = dict[str | None, dict[str, TExpr[T]]]
 
 
+class CompilationFlags(Flag):
+    """A collection of compilation flags to modify `compile_routine` functionality."""
+
+    EXPAND_RESOURCES = auto()
```
mstechly, line 85: Have you checked how this displays in the docs?

src/bartiq/compilation/_compile.py
```diff
@@ -274,7 +294,7 @@ def _compile(
     inputs: dict[str, TExpr[T]],
     context: Context,
     derived_resources: Iterable[DerivedResources] = (),
-    allow_transitive_resources: bool = True,
+    compilation_flags: CompilationFlags = CompilationFlags(0),
```
mstechly, line 297: To me `CompilationFlags` doesn't really say much. Can we perhaps add a comment explaining what it means?
Or just have `None` as default value?

tests/compilation/test_repetitions.py
```diff
@@ -264,14 +265,15 @@ def test_custom_sequence_throws_error_when_replacing_iterator_symbol_on_evaluate
...
-    compiled_routine = compile_routine(routine, allow_transitive_resources=transitive_resources).routine
+    compiled_routine = compile_routine(routine, compilation_flags=compilation_flags).routine
     with pytest.raises(BartiqCompilationError, match=r"Tried to replace symbol that's used as iterator symbol*"):
         evaluate(compiled_routine, assignments).routine
 
 
-# At present, allow_transitive_resources=True does not raise an error when attempting to compile a routine
-# where a resource value has the same symbol as an iterator symbol. This is because the resource hierarchy is not
-# compiled fully, and so the parameter map during the compile sequence does not contain child information.
+# At present, the default method of compilation (with transitive resourceSs does not raise an error
```
mstechly, line 273: `resourceSs` typo.

src/bartiq/compilation/_compile.py
Brendan-Reid1991: I didn't expose `CompilationFlags` at the top level, and `bartiq.compilation` is not included in the docs!  I added it in, and it displays... reasonably. 

Since the `compilation` folder is public and has public scripts, do you want me to add that addition to the docs onto this PR?
Screenshot 2025-05-14 at 11 50 53

src/bartiq/compilation/_compile.py
Brendan-Reid1991: Do you mean name-wise? The reason the default is not `None` here is to avoid code duplication, e.g. on line 138 the default is `None` because it is public, and and this is accounted for on line 157
```compilation_flags=compilation_flags or CompilationsFlags(0)```
where `CompilationFlags(0)` is the null value. I didn't want the exact same line to be in the private `_compile` function, so I set the default as the null value rather than `None` to satisfy the mypy checks!

src/bartiq/compilation/_compile.py
mstechly: Ok great!
Yes please 🙏

mstechly (approved):

PR #202, opened 2025-05-12
chore: add resource_values property to BaseRoutine

pqvr:
- Added `resource_values` to `BaseRoutine` as a propety
- Added unit test for `resource_values`.

### DIFF
--- a/.github/workflows/lint_pr_title.yaml
+++ b/.github/workflows/lint_pr_title.yaml
@@ -21,19 +21,10 @@ jobs:
       - name: Install commitlint
         run: npm install @commitlint/cli @commitlint/config-conventional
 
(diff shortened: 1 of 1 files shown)


dexter2206 (approved):

mstechly:
@pqvr have you checked whether there are any places where we can use this replacement? I'd expect sth in the `latex` module perhaps?

mstechly:
@pqvr I looked into `latex` and potentially it could be changed in places like `_format_resources` and `_get_resources_lines`. But that would be an overkill. Couldn't find anything else that's obvious 🤷‍♂️ 
I agree, external projects are higher priority 👌

mstechly (approved):

Issue #163, opened 2025-01-29
Fix displaying resource name `local_highwater`.


mstechly:
## Describe the bug
When we have a resource called `local_highwater`, when we export it to latex, for some reason the `cal` part gets cut.

<img width="158" alt="Image" src="https://github.com/user-attachments/assets/f2cdcb20-0442-4f14-bbcf-9b22c391bd68" />

I've also seen case where it was wrong in some other way, but I couldn't reproduce it. Though it was still related to the fact that there's `cal` in the name, which probably needs to get escaped/sanitized somehow.

### How to reproduce

Run the following in a jupyter notebook:

```
from qref import SchemaV1
from bartiq import compile_routine
from bartiq.integrations import routine_to_latex
from IPython.display import Math

qref_dict = {
    "name": "my_routine",
    "type": None,
    "resources": [
        {"name": "local_highwater", "type": "other", "value": "10"},
    ],
}

alias_sampling_qref = {"version": "v1", "program": qref_dict}
uncompiled_routine = SchemaV1(**alias_sampling_qref)
compiled_routine = compile_routine(uncompiled_routine)

Math(routine_to_latex(compiled_routine.to_qref()))
```

The `routine_to_latex` function is contained [here](https://github.com/PsiQ/bartiq/blob/main/src/bartiq/integrations/latex.py)

PR #208, opened 2025-05-28
fix: avoid SymPy ``LatexPrinter`` modifiers to change subscripted parameter symbols

antalszava:
## Description

This PR is meant to close https://github.com/PsiQ/bartiq/issues/163.

**Context**

The rendering issue arises when calling the ``symbols`` SymPy function, which under the hood attempts to create SymPy ``Symbol`` objects and then, once done, prints them.

image

A ``Symbol`` object in SymPy is also a ``Printable`` object, whose [``_repr_latex_`` method](https://github.com/sympy/sympy/blob/da6d1f3cb324ad0bf74d123d85c390029be997ce/sympy/core/_print_helpers.py#L54) is being called in a Jupyter environment. However, this method defaults to a behaviour that interprets certain input suffixes as default modifiers. The example string from #163 had the specific `cal` suffix, which happens to be one of the modifiers ([see here](https://github.com/sympy/sympy/blob/da6d1f3cb324ad0bf74d123d85c390029be997ce/sympy/printing/latex.py#L89) for the modifier dictionary that is being used in the ``LatexPrinter`` class).

**Implemented solution**

The implemented solution does the following:

* In case we have a subscript expression (such as in issue #163), we check if the input string ends in either of the modifiers from SymPy's modifier dictionary;
* If the input string indeed ends in such a modifier, then we enclose the string with curly brackets such that it is made into a LaTeX grouping expression, effectively avoiding SymPy's logic as well as preserving the expected Latex rendering.

*Note:* no dependencies were added in the solution.

- [X] I have included test cases validating introduced feature/fix.
*Note*: At the moment, a single test case was added. I'm happy to include more testing if that helps further this bug fix. :ok_hand: 
*Note*: Having checked the documentation, as `routine_to_latex` is not explicitly mentioned in the docs, no changes were made. Having said that, I'm happy to include any relevant updates, if that helps. :+1:

### DIFF
--- a/src/bartiq/integrations/latex.py
+++ b/src/bartiq/integrations/latex.py
@@ -24,6 +24,7 @@ from qref.schema_v1 import (
     SchemaV1,
 )
 from sympy import latex, symbols
+from sympy.printing.latex import modifier_dict as sympy_latex_modifier_dict
 
 from ..symbolics.sympy_backend import parse_to_sympy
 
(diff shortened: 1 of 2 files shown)


mstechly (approved):

mstechly:
@antalszava thank you for the PR, everything looks good!
Could you push an empty commit to trigger `cla-bot` again?
Unfortunately I don't see an option to restart that job manually :(

PR #210, opened 2025-05-28
feat(transform): add add_circuit_volume feature, tests, and notebook docs

### LINKED ISSUE #150: Add function for calculating `circuit_volume`
## Is your feature request related to a problem? If so, please describe it.
Circuit volume is a quantity that's often useful and would be good to be able to automatically add it into routines.

## How you envision the solution

Calculating it is basically doing: `circuit_volume = aggregated_t_gates * qubit_highwater` for each subroutine.

I think that within the scope of this issue, we should add to [`transform` module](https://github.com/PsiQ/bartiq/blob/main/src/bartiq/transform.py) an `add_circuit_volume` method with the following signature:

```
def add_circuit_volume(
    routine: CompiledRoutine[T],
    name_of_aggregated_t: str = "aggregated_t_gates",
    backend: SymbolicBackend[T] = BACKEND,
) -> CompiledRoutine[T]:
```

Calculating circuit volume would depend on `aggregated_t_gates`, which most routines would probably not have right away. `aggregated_t_gates` is not something that's defined anywhere in Bartiq. It's just a placeholder for the number of T gates coming from the fact some "bigger" gates, such as Toffolis or rotations, which can be decomposed into T gates. 

Therefore, we shouldn't worry too much about adding the fact that `aggregated_t_gates`, since the user can add it with `add_aggregated_resources` if the wish.In the tests we should just have a routine that already has a resource named appropriately.

poig:
## Description

this pr aim to close #150

This PR introduces the ``add_circuit_volume`` utility to the transform module. Circuit volume is a useful composite metric in quantum resource estimation, defined as the product of the total T-gate count (after aggregation) and the qubit high-water mark. This feature is needed to enable users to easily and automatically add a ``circuit_volume`` resource to their routines, which is commonly used for benchmarking and hardware feasibility analysis.

Implemented Solution:

Added the ``add_circuit_volume`` function to transform.py. This function traverses a compiled routine and adds a ``circuit_volume`` resource to each subroutine where both ``aggregated_t_gates`` and ``qubit_highwater`` are present.
The function strictly uses ``aggregated_t_gates`` as the T-gate resource, following the original design intent. No fallback to ``T_gates`` is performed, ensuring users explicitly aggregate resources if needed.
Comprehensive test cases were added to test_transform.py to validate the feature, including numeric, symbolic, and missing-resource scenarios.
Documentation tutorial notebook were updated to demonstrate usage and clarify the requirements for using ``add_circuit_volume``.

### DIFF
--- a/docs/tutorials/01_basic_example.ipynb
+++ b/docs/tutorials/01_basic_example.ipynb
@@ -708,6 +708,46 @@
   },
   {
    "cell_type": "markdown",
+   "id": "b9394ebb",
+   "metadata": {},
+   "source": [
+    "Now that we have evaluated our routine, we can automatically add a `circuit_volume` resource. This requires that both `aggregated_t_gates` and `qubit_highwater` resources are present. If your routine uses `T_gates`, you can set `aggregated_t_gates` to `T_gates` for this purpose."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 16,
+   "id": "d81a2f09",
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "Circuit volume: 3665\n"
+     ]
+    }
+   ],
+   "source": [
+    "from bartiq import Resource, ResourceType\n",
+    "from bartiq.transform import add_circuit_volume\n",
+    "\n",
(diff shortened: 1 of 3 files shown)


Brendan-Reid1991 (changes requested):
Only a few comments to address, then this would be ready to merge once the CLA registers.

PR #212, opened 2025-05-28
feat(symbolics): improve nlz and round error handling for non-integer and non-SymPy object input

### LINKED ISSUE #180: Passing a non-integer into nlz results in unhelpful error message.
## Describe the bug
When evaluating large expressions that contain [nlz](https://github.com/PsiQ/bartiq/blob/main/src/bartiq/symbolics/sympy_interpreter.py#L141), if you are not careful with your datatypes you can get an unhelpful error message: 
```
    TypeError: Expected a number but got Add/Mul/Pow
```

The issue comes from a float-type representing an integer, i.e. `10.0` vs `10`.

This can occur as part of a larger workflow where the variables are not handled directly, are instead being parsed from elsewhere and then cast to specific types. For instance, it is not immediately obvious that using `10.0` should produce quantitatively or qualitatively different results from using `10` -  `sympy` is a maths package after all! As the error message is being handled by `sympy`, this can be difficult to debug.

## How to Reproduce
```
    expr = parse_to_sympy("B*nlz(A) + C")
    sympy_backend.substitute(expr, replacements={"A": 10.0, "B": 12, "C": 9})
```

## Expected behavior
I think an error should be raised if a non-integer value is passed to `nlz`. I toyed with the idea of doing some checks on the input, i.e. are there any values after the decimal point (is it a real float or an integer-as-float), but this would incur overhead in performance and is tantalising close to correcting user error. It's not _really_ a user error to input `10.0` vs `10`, but I can't think of a better way to handle it. 

## Actual behavior
Big ol' `TypeError`, depending on the type of expression input. 

## Environment:**
 - OS: OSX
 - Python version: 3.12.3

poig:
## Description

this pr aim to close #180

This PR improves the ``nlz`` symbolic function in ``sympy_interpreter.py`` to ensure it raises clear and specific errors for all non-integer input types, as required for correct symbolic and numeric behavior in Bartiq's sympy backend.

Context:
Previously, the ``nlz`` function did not consistently raise errors for non-integer input, which could lead to silent failures or incorrect symbolic evaluation, especially when users provided floats or symbolic (non-integer) values. This was highlighted by new parameterized tests that expect ``TypeError`` with specific messages for floats and generic messages for other non-integer types.

Updated the nlz function to:
- ``Raise TypeError("Input to nlz must be an integer, not a float")`` for any float input (Python or sympy Float).
- ``Raise TypeError("Input to nlz must be an integer")`` for all other non-integer types (including symbolic).
- This ensures all test cases in ``test_sympy_interpreter.py`` pass and that error messages are clear and actionable for users.

### DIFF
--- a/tests/symbolics/test_sympy_interpreter.py
+++ b/tests/symbolics/test_sympy_interpreter.py
@@ -13,18 +13,65 @@ Tests for the sympy interpreter.
 import string
 
 import pytest
-from sympy import (Float, Function, Heaviside, Integer, LambertW, Min, Mod,
-                   Product, Rational, Sum, Symbol, acos, acosh, acot, acoth,
-                   acsc, acsch, asec, asech, asin, asinh, atan, atanh, cbrt,
-                   ceiling, cos, cosh, cot, coth, csc, csch, exp, floor, frac,
-                   gamma, im, log, prod, re, sec, sech, sin, sinh, sqrt, tan,
-                   tanh)
+from sympy import (
+    Float,
+    Function,
+    Heaviside,
+    Integer,
+    LambertW,
+    Min,
+    Mod,
+    Product,
+    Rational,
+    Sum,
+    Symbol,
+    acos,
+    acosh,
+    acot,
+    acoth,
+    acsc,
+    acsch,
+    asec,
+    asech,
+    asin,
+    asinh,
+    atan,
+    atanh,
+    cbrt,
+    ceiling,
+    cos,
+    cosh,
+    cot,
+    coth,
+    csc,
+    csch,
+    exp,
+    floor,
+    frac,
+    gamma,
+    im,
+    log,
+    prod,
+    re,
+    sec,
+    sech,
+    sin,
+    sinh,
+    sqrt,
+    tan,
+    tanh,
+)
 from sympy.codegen.cfunctions import exp2, log2, log10
 from sympy.core.numbers import S as sympy_constants
 
 from bartiq.symbolics.sympy_backend import parse_to_sympy
-from bartiq.symbolics.sympy_interpreter import (SPECIAL_PARAMS, Max, Round,
-                                                multiplicity, nlz)
+from bartiq.symbolics.sympy_interpreter import (
+    SPECIAL_PARAMS,
+    Max,
+    Round,
+    multiplicity,
+    nlz,
+)
 from bartiq.symbolics.sympy_serializer import serialize_expression
 
 


src/bartiq/symbolics/sympy_interpreter.py
```diff
@@ -142,9 +142,15 @@ def eval(cls, p, n):
 class nlz(Function):
     @classmethod
     def eval(cls, n):
+        from sympy import Float, Integer
```
mstechly, line 145: issue: do not use imports inside function's body.  Use imports at the top of the file.
Also, in this case we're already importing both `Float` and `Integer` at the top of the file, so this is unnecessary.

tests/symbolics/test_sympy_interpreter.py
```diff
@@ -397,3 +397,20 @@ def test_sympy_interpreter_warns_about_using_caret_sign_for_exponentiation():
     expr = "x ^ 2"
     with pytest.warns(match=r"Using \^ operator to denote exponentiation is deprecated\."):
         _ = parse_to_sympy(expr)
+
+
+@pytest.mark.parametrize(
+    "value,expected,raises,match",
+    [
+        (8, 3, None, None),  # integer input, should succeed
+        (8.0, None, TypeError, "must be an integer, not a float"),  # float integer, should raise
```
mstechly, line 406: I think if we implement the suggestion I made above we could just get rid of `match`, WDYT?

mstechly (changes requested):

tests/symbolics/test_sympy_interpreter.py
poig: thanks for reviewing, I just update it with the idea, take a look

poig:
should we make it as a warper instead like this:
```
class nlz(Function):
    @sympy_numeric_eval
    def eval(cls, n):
        # Only called if n.is_number is True
        if not n.is_integer:
            raise TypeError(...)
        # ...rest of logic...

class multiplicity(Function):
    @sympy_numeric_eval
    def eval(cls, p, n):
        # Only called if p.is_number and n.is_number are True
        # ...rest of logic...
```

poig:
should we also modify Round.eval
```
class Round(Function):
    """Defines a delayed-evaluation version of Round."""

    @classmethod
    def eval(cls, x, ndigits=None):
        # Only evaluate if x is a number
        if not getattr(x, "is_number", False):
            return None  # symbolic x, leave unevaluated

        if ndigits is None:
            return round(float(x))
        if getattr(ndigits, "is_number", False):
            if getattr(ndigits, "is_integer", None) is True:
                return round(float(x), int(ndigits))
            raise TypeError(f"Input ndigits must be an integer; found {ndigits}")
        if isinstance(ndigits, int):
            return round(float(x), ndigits)
        if isinstance(ndigits, (float, str)):
            raise TypeError(f"Input ndigits must be an integer; found {ndigits}")
        return None  # symbolic ndigits
```

src/bartiq/symbolics/sympy_interpreter.py
```diff
@@ -101,49 +101,68 @@ class Round(Function):
 
     @classmethod
     def eval(cls, x, ndigits=None):
-        """Define the function's immediate evaluation in the case where the input is a number."""
-        # Evaluate for explicit number x
-        if x.is_number:
+        # Only evaluate if x is a number
```
mstechly, line 104: Why did you touch this class at all? I think it worked just fine and now the logic seems overly complicated, so I'm curious if there was some specific reason for that?

mstechly (changes requested):
I don't think updating the `Round` function is out of scope and not needed for this PR. Once you revert these changes, I'm happy with it.

mstechly (approved):

mstechly:
@poig Please update to `main` and fix that one `isort` check and we're good to go :)

poig:
> @poig Please update to `main` and fix that one `isort` check and we're good to go :)

sorry I am confuse since I am using the same version of yaml file 5.13.2, but still can't find the import order issue, it just reformat the whole thing

mstechly:
Try with the following settings: `isort --profile black --filter-files` .

(source: https://github.com/PsiQ/bartiq/blob/a3d9da4b21ffdc5bf1cb10a9e567f1585e54ad98/.pre-commit-config.yaml#L25)

poig:
@mstechly Thank you I learn a lot!

PR #215, opened 2025-05-28
refactor bartiq to use single symbol for expressing pow method

### LINKED ISSUE #138: refactor bartiq to use single symbol for expressing `pow` method

SumbulZK:
## Description

Currently in Bartiq we allow to express pow function as both ^ and **.
This is problematic since:

^ in Python is a XOR operator,
an expression including one symbol could be later serialized and the symbol might turn into a different one, which could be confusing.
Refactored code to use ** instead of ^

Closes #138 

- [ ] I have updated documentation. (could not find instances using ^)

### DIFF
--- a/src/bartiq/symbolics/ast_parser.py
+++ b/src/bartiq/symbolics/ast_parser.py
@@ -161,7 +161,7 @@ def _replace_xor_op(expression: str) -> str:
 
 
 # Preprocessing stage replacing ^ with ** for exponentiation
-_XOR_REPLACEMENT = _PreprocessingStage(matches=_contains_xor_op, preprocess=_replace_xor_op)
+_XOR_OP_REPLACEMENT = _PreprocessingStage(matches=_contains_xor_op, preprocess=_replace_xor_op)
 
 
 # Sequence of all known preprocessing stages.
@@ -173,7 +173,7 @@ _PREPROCESSING_STAGES = (
     _PORT_REPLACEMENT,
     _LAMBDA_REPLACEMENT,
     _IN_REPLACEMENT,
-    _XOR_REPLACEMENT,
+    _XOR_OP_REPLACEMENT,
 )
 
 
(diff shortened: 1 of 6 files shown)


src/bartiq/symbolics/ast_parser.py
```diff
@@ -164,6 +150,20 @@ def _replace_in(expression: str) -> str:
 # Preprocessing stage replacing "in"s with _in
 _IN_REPLACEMENT = _PreprocessingStage(matches=_contains_in, preprocess=_replace_in)
 
+
```
mstechly, line 153: I think we can just get rid of these now.

src/bartiq/symbolics/ast_parser.py
mstechly, line 1: You also should update docstring in `_PreprocessingStage`

src/bartiq/symbolics/sympy_interpreter.py
mstechly, line 1: `test_sympy_interpreter_warns_about_using_caret_sign_for_exponentiation` will start failing if you remove `_XOR_OP_REPLACEMENT` from `ast_parser`.
I think that wanting to use `^` symbolically is a reasonable thing, but we're not supporting it right now.
Please change this test to check `pytest.raises` and edit logic in `ast_parser.py` to raise `NotImplementedError` when we encounter `ast.BitXor`.

Alternatively, feel free to play with it a bit more and see if you can add support for `^` to the sympy code :)

mstechly (changes requested):

SumbulZK:
@mstechly I know, I just wanted to contribute 😄

src/bartiq/symbolics/sympy_interpreter.py
SumbulZK: @mstechly 
That makes sense, I will edit the logic to show XOR is not implemented yet to close this issue, and then work separately on implementing support for ^

SumbulZK: Thanks for the feedback, I have added the requested changes

mstechly:
@SumbulZK do you have any intention to fix the pipeline and finalize this PR, or should I close it?

SumbulZK:
@mstechly Apologies, for the delay. I will fix the pipeline and close this within the week

PR #216, opened 2025-05-28
Further improve visualisations

antalszava:
## Description

Include description of feature this PR introduces or a bug that it fixes. Include the following information:

- Context: why is it needed? In case of bug, what was the cause for the bug?
- Concise description of the implemented solution.
- If any dependencies are added, list them and justify why they are needed.

### DIFF
--- a/src/bartiq/visualisations.py
+++ b/src/bartiq/visualisations.py
@@ -95,7 +95,7 @@ class TreeMap:
 ######################################################
 
 
-type Contributions = dict[str, Number]
+Contributions = dict[str, Number]
 
 
 def _get_child_contributions(routine: CompiledRoutine, resource: str, scale_to: str | int = 1) -> Contributions:
@@ -109,7 +109,7 @@ def _get_child_contributions(routine: CompiledRoutine, resource: str, scale_to: 
     }
 
 
-type NestedContributions = tuple[Contributions, dict[str, NestedContributions]]
+NestedContributions = tuple[Contributions, dict[str, NestedContributions]]
 
 
 def _get_descendant_contributions(
--- a/src/bartiq/visualisations.py
+++ b/src/bartiq/visualisations.py
@@ -109,7 +109,7 @@ def _get_child_contributions(routine: CompiledRoutine, resource: str, scale_to: 
     }
 
 
-NestedContributions = tuple[Contributions, dict[str, NestedContributions]]
+NestedContributions = tuple[Contributions, dict[str, Contributions]]
 
 
 def _get_descendant_contributions(


Brendan-Reid1991:
Before I start my review, can you:
a) Sync your fork to the PsiQuantum bartiq repo
b) Change the target branch of this PR to `main` rather than `improve_visualisations`

That'll give me a better idea of the changes!

src/bartiq/symbolics/symbolic_manipulator.py
Brendan-Reid1991, line 1: Delete this file! This is a remnant of another branch that `improve_visualisations` branched off of.

src/bartiq/symbolics/test.ipynb
Brendan-Reid1991, line 1: Delete this file

src/bartiq/visualisations.py
Brendan-Reid1991, line 1: Change name to `visualizations.py`.

src/bartiq/visualisations.py
```diff
@@ -0,0 +1,191 @@
+from __future__ import annotations
```
Brendan-Reid1991, line 1: Module level docstring

src/bartiq/visualisations.py
```diff
@@ -0,0 +1,191 @@
...
+    Raises:
+        ValueError: If any resource in the top level routine has a non-numeric value.
+    """
+
+    COLUMNS = ["Routine", "Parent", "Contribution"]
+
+    def __init__(self, routine: CompiledRoutine):
+        self.routine = routine
+        if any(not isinstance(x.value, Number) for x in self.routine.resources.values()):
+            raise ValueError(f"{self.__class__.__name__} only accepts numeric routines.")
```
Brendan-Reid1991, line 28: I think I wrote this test, but it doesn't work. Change to:

```suggestion
        if not all(isinstance(x.value, Number) for x in self.routine.resources_values().values():
            raise ValueError(f"{self.__class__.__name__} only accepts numeric routines.")
```

We will also need a test to make sure this exception is thrown!

src/bartiq/visualisations.py
```diff
@@ -0,0 +1,191 @@
...
+
+        self.valid_resources = set([resource_name for resource_name in self.routine.resources.keys()])
+
+    def get_dataframe(self, resource: str, scale_to: Literal["parent", "root"] | Number = 1) -> pd.DataFrame:
```
Brendan-Reid1991, line 32: I think remove the option to arbitrarily scale it with a `Number`. I think it has limited utility and could only cause confusion.

src/bartiq/symbolics/sympy_backend.py
```diff
@@ -190,8 +190,11 @@ def substitute(
         functions_map: Mapping[str, Callable[[TExpr[Expr]], TExpr[Expr]]] | None = None,
     ) -> TExpr[Expr]:
 
-        symbols_in_expr = self.free_symbols_in(expr)
-        restricted_replacements = [(symbols(old), new) for old, new in replacements.items() if old in symbols_in_expr]
+        restricted_replacements = [
+            (old_symbol, replacement)
+            for old_name, replacement in replacements.items()
+            if (old_symbol := _symbol_by_name(expr, old_name))
+        ]
```
Brendan-Reid1991, line 197: Delete the changes in this file.

src/bartiq/visualisations.py
```diff
@@ -0,0 +1,191 @@
...
+                    )
+                if child in grandchildren_contributions:
+                    df = _update_dataframe_recursive(grandchildren_contributions[child], parent=child, df=df)
+            return df
+
+        return _update_dataframe_recursive(output_from_parent, self.routine.name, df)
+
+    @classmethod
+    def dataframe_with_unique_routine_names(cls, df):
+        """Get the dataframe with unique routine names ready to be plotted.
```
Brendan-Reid1991, line 71: No need for this to be a class method, you can make it a private function at the bottom of this file!

src/bartiq/visualisations.py
```diff
@@ -0,0 +1,191 @@
...
+
```
Brendan-Reid1991, line 72: Amazing docstring, thank you!

src/bartiq/visualisations.py
```diff
@@ -0,0 +1,191 @@
...
+        Returns:
+            pd.DataFrame
+        """
+        routine_name_counter: dict[str, int] = {}
+        old_name_to_latest_unique_name: dict[str, str] = {}
+        result = []
+
+        routine_col = cls.COLUMNS[0]
+        parent_col = cls.COLUMNS[1]
```
Brendan-Reid1991, line 108: When removing this function from the class to be a standalone, private, function. Access these with `TreeMap.COLUMNS`. No need for instantiation to access class variables.

docs/tutorials/03_advanced_examples.ipynb
Brendan-Reid1991, line 1: Remove this change just by `git checkout main docs/tutorials/03_advanced_examples.ipynb`

src/bartiq/symbolics/sympy_backend.py
Brendan-Reid1991: ** via `git checkout main src/bartiq/symbolics/sympy_backend.py`

src/bartiq/visualisations.py
```diff
@@ -0,0 +1,191 @@
...
+
+        Args:
+            resource: Resource type to isolate.
+            scale_to: Optional input to rescale the values. Accepts a numeric value or string literal `parent`,
+            which will scale each resource count as a fraction of its parents. Defaults to 1, which is no scaling.
+
+        Returns:
+            pd.DataFrame
+        """
+        df = pd.DataFrame([], columns=self.COLUMNS)
```
Brendan-Reid1991, line 43: Add check that passed `resource` is in the `self.valid_resources` list, raise an Exception if not.

Brendan-Reid1991:
OH, btw, to pass the quality checks you'll need to run:
```
black -l 120 .
mypy --install-types --non-interactive src
isort --profile=black .
```

Brendan-Reid1991:
@antalszava Thank you! This looks great. 

The one thing I'm unsure about is the new methods on `Routine` objects. Since a `Routine` object is distinct from a `CompiledRoutine`, I'm not sure it is kosher to accept `Routine` and then compile it behind the scenes. Since a tree map is a) nested and b) numeric by design, accepting:
- a numeric, _uncompiled routine_ violates a), since the proper relationships and have not been determined (and it is not the job of the tree map class to compile!)
- a symbolic, _uncompiled routine_ violates a) and b), and a treemap of symbolic equations would be nigh impossible to represent cleanly. 
- a symbolic _compiled routine_ violates b), and we use the same reasoning as above
- a numeric _compiled routine_ doesn't violate either of these, and should be the only acceptable input. 

Furthermore, while I appreciate the forethought of testing all the children and grandchildren for numerics, it is unnecessary if the input to `TreeMap` is type hinted as only a `CompiledRoutine`. The compilation procedure in `bartiq` ensures that that nth level inherits from the (n+1)th level, where 0th is the root, 1st is the first layer of children etc. So, a `CompiledRoutine` object with numeric top level resources and symbolic lower level resources should not be possible. I know this to be true, but whether or not we have thought of all possible "unintended ways of using bartiq" and tried to catch them, I can't say**. But this would not be the job of `TreeMap` to catch.

By only accepting _numeric, CompiledRoutines_, we also avoid the issue of the `child_name.resource_name` format of the unevaluated resource values. Essentially, the user _has_ to evaluate their routine to be accepted by `TreeMap`. It will be enough to type hint for `CompiledRoutine`, and then do a check on the top level resources to ensure they are all `Number` types. You've done more work than you needed to! But I appreciate it. 

I can merge this today, with the following changes:
- Remove new methods from `_routine.py`
- Change type hinting for `TreeMap` to only accept `CompiledRoutines`, and check for all numeric top level resources.
- Move from `if resource not in self.valid_resources` check from `plot` into `get_dataframe`, as `get_dataframe` is also public, and `plot` calls `get_dataframe`. So it only needs to exist in `get_dataframe`!
- Line 127 in `visualizations.py` should be changed from `"Contribution" -> contribution` to avoid string duplication.

Thank you again!


**NB: We did not intend `Routine`/`CompiledRoutine` objects to be manually created, so there may be a way to strong-arm it. This falls under "unintended ways of using bartiq" though.

Brendan-Reid1991 (approved):

Brendan-Reid1991:
Amazing @antalszava ! I've approved there now, once the pipeline passes I'll ping Michal to merge and make sure you get your bounty :)

Issue #220, opened 2025-05-30
Comparative code coverage reports


pqvr:
Related to: https://github.com/PsiQ/bartiq/pull/94

Obtain a report that provides extensive coverage has changes in a given PR. In the related PR (https://github.com/PsiQ/bartiq/pull/94), configuring was done with `pycobertura`, but it just shows coverage, not a change in coverage. This issue is to enable the showing of the change in coverage as well.

PR #221, opened 2025-05-30
feat(tests): code cov with cov change comp

pqvr:
## Description

Closes: #220 

Workflow to track coverage from main to current PR.


tests/compilation/test_compile.py
```diff
@@ -234,7 +234,7 @@ def test_compilation_fails_if_input_ports_has_size_depending_on_undefined_variab
 # mess up the parser because of the recursion limit.
 # Also note: these strings are specific to sympy.
 @pytest.mark.order(-1)
-@pytest.mark.timeout(30)
+@pytest.mark.timeout(300)
```
mstechly, line 237: One moment here!
30 was set here to be a safe buffer. If this is running over 30 seconds, then we might have a performance problem.
Why did you change it?

mstechly (commented):
I don't think that we need to show a full `Code Coverage Summary` every time, as this is not particularly useful info. Can we show just the diff?

tests/compilation/test_compile.py
mstechly: But it does pass on other jobs on CI.
So it looks like generating coverage increases the runtime that significantly?
Can you check the logs & run it locally to see what are the time differences with & without coverage?

pqvr: It doesn't appear to increase significantly from what I can tell from prior jobs and workflows. It averages about three minutes either with or without the coverage reports. I think the jump from `30` -> `300` is likely generous, and while it doesn't need a delay with that long of a duration, it does buffer a bit more just in case. I can ratchet down the number if if you feel like 300 is too high, I think it just needs to be somewhere north of 30 to prevent timeouts.

mstechly: Sounds good for me!

pqvr: Good point, @dexter2206. 

I've gone ahead and added a `no-perf-tests` function in the `conftest.py` and added the `--no-perf-tests` flag to skip over these in the coverage. Thank you, @mstechly and @dexter2206 !

PR #226, opened 2025-06-09
refactor: made pandas and plotly optional installs

Brendan-Reid1991:
## Description

The addition of pandas and plotly in PR #216 made the install noticeably slower.
I've added a new extras group in the .toml `visualization` to make the tree map functionality essentially optional, similar to the optimisation functionality.


pyproject.toml
```diff
@@ -21,10 +19,13 @@ ipytree = { version = "^0.2.2", optional = true }
 ipywidgets = { version = "^8.1.2", optional = true }
 traitlets = { version = "^5.14.3", optional = true }
 scipy = { version = "^1.14.1", optional = true }
+pandas = { version = "^2.2", optional = true }
+plotly = { version = "^5.17", optional = true }
 
 [tool.poetry.extras]
 jupyter = ["ipytree", "ipywidgets", "traitlets"]
 optimization = ["scipy"]
+visualization = ["pandas", "plotly"]
```
mstechly, line 28: @Brendan-Reid1991 How about having just one extra `interactive`, that would combine `jupyter` and `visualization`?
I understand that visualizations are not necessary part of the interactive workflow, but that's probably how 90% of people will be using it and it might be just better API.
WDYT?

src/bartiq/analysis.py
```diff
@@ -11,6 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+# mypy: disable-error-code="arg-type,misc"
```
mstechly, line 14: @dexter2206  :) ?

tests/test_visualizations.py
```diff
@@ -13,20 +13,29 @@
...
 from bartiq import Routine, compile_routine
 from bartiq.symbolics.sympy_backend import SympyBackend
-from bartiq.visualizations import TreeMap, _dataframe_with_unique_routine_names
 
+try:
+    import pandas as pd
 
+    from bartiq.visualizations import TreeMap, _dataframe_with_unique_routine_names
+
+    SKIP_VIS_TESTS = False
```
mstechly, line 27: [comment] I'd personally go with `viz`, as that's more how you spell it. I'm not asking to change it, but actually, curious what are your thoughts on this ;)

mstechly (approved):

src/bartiq/analysis.py
Brendan-Reid1991: If @dexter2206 can figure out why this file fails mypy on py3.10, I would appreciate it!

To reproduce (with this line in analysis.py deleted):

## _On `main`_:
```
poetry env use 3.10
poetry install
poetry run mypy --install-types --non-interactive src
```
will **pass**. 

## _On this branch_:
```
poetry install ## I use py3.12
poetry run mypy --install-types --non-interactive src
```
will **pass**.

```
poetry env use 3.10
poetry install
poetry run mypy --install-types --non-interactive src
```
will **fail**.

src/bartiq/visualizations.py
```diff
@@ -18,9 +18,15 @@
...
-from plotly.graph_objs._figure import Figure as PlotlyFig  # type: ignore
+try:
+    import pandas as pd
+    import plotly.express as px  # type: ignore
+    from plotly.graph_objs._figure import Figure as PlotlyFig  # type: ignore
+except ModuleNotFoundError:
+    raise ModuleNotFoundError(
+        "To use the Bartiq visualization functionality, reinstall with `poetry install -E visualization`."
+    )
+
```
dexter2206, line 29: **question**: Given that this error message is probably intended to be seen by the end users (because the developers know how to install extras), shouldn't we actually state that bartiq should be reinstalled with `pip install "bartiq[visualization]"`? Also we could think about using the `raise ... from ...` construct to obtain a slightly better looking traceback.

```suggestion
except ModuleNotFoundError as e:
    raise ModuleNotFoundError(
        'To use the Bartiq visualization functionality, reinstall with `pip install "bartiq[visualization]"`.`
     ) from e

```

**praise**: I think adding this error message will be really helpful for the users. Great idea!

tests/test_visualizations.py
```diff
@@ -13,20 +13,29 @@
...
+    import pandas as pd
 
+    from bartiq.visualizations import TreeMap, _dataframe_with_unique_routine_names
+
+    SKIP_VIZ_TESTS = False
+except ModuleNotFoundError:
+    SKIP_VIZ_TESTS = True
+
+
+@pytest.mark.skipif(SKIP_VIZ_TESTS, reason="Requires pandas and plotly to test.")
```
dexter2206, line 32: **question**: Have you considered using `importorskip` from pytest? Check out the docs [here](https://docs.pytest.org/en/stable/how-to/skipping.html#skipping-on-a-missing-import-dependency). Very rarely used feature (I learned about it only recently), but it gives you a very clean whole-module skip on missing import.

PR #227, opened 2025-06-09
chore: update poetry.lock file for dependabot

pqvr:
## Description

Resolving the (closed) dependabot issues in one MR (which was done by regenerating the `poetry.lock` file from scratch:
- https://github.com/PsiQ/bartiq/pull/222
- https://github.com/PsiQ/bartiq/pull/223
- https://github.com/PsiQ/bartiq/pull/224


pyproject.toml
```diff
@@ -14,6 +14,7 @@ python = "^3.10"
 pydantic = "^2.7"
 sympy = "^1.12"
 qref = "^0.11.0"
+h11 = "^0.16.0"
```
mstechly, line 17: I don't like the idea of adding `h11` as an explicit dependency, since it's a transient dependency, we don't use it anywhere directly.
At least there should be a comment on why it's there.

mstechly: I think the tree failed for you as you already had `h11` specified in pyproject.toml.

When I ran it on a fresh repo I got:

```
├── jupyterlab >=4.3.6,<4.4
│   ├── async-lru >=1.0.0
│   │   └── typing-extensions >=4.0.0
│   ├── httpx >=0.25.0
│   │   ├── anyio *
│   │   │   ├── exceptiongroup >=1.0.2
│   │   │   ├── idna >=2.8
│   │   │   ├── sniffio >=1.1
│   │   │   └── typing-extensions >=4.5 (circular dependency aborted here)
│   │   ├── certifi *
│   │   ├── httpcore ==1.*
│   │   │   ├── certifi * (circular dependency aborted here)
│   │   │   └── h11 >=0.13,<0.15
```

So I went down that rabbithole and it looks like it's already fixed in `httpcore` (see this MR: https://github.com/encode/httpcore/pull/1008), but it just has not been released yet.

So I would go with option 2, the problem will solve itself once they release it and we update poetry lock.

Issue #228, opened 2025-06-13
Compilation engine vs User-facing API


Brendan-Reid1991:
After discussions with @dexter2206 at the onsite in June 2025, it was suggested that maybe we should begin to separate the parts of Bartiq that are the compilation engine, and the parts of Bartiq that a user will want to use/abuse. 

With a `CompiledRoutine` object, a user will want to do multiple things with it (`evaluate`, traverse the DAG, determine which resources have unknown functions in them, etc). We can (and have begun to) add helper functions to the `Routine` and `CompiledRoutine` objects to help in this aspect. Maybe the simplest example of this is the new `resource_values` attribute, which we implemented because we found that users would use it more. 

As another example, consider the `find_descendants` feature introduced during Unitary Hack ([link](https://github.com/PsiQ/bartiq/blob/0bd387aa1e5c7d3afb470c9e6950eea974f576db/src/bartiq/_routine.py#L165)). This functionality was introduced to address a specific problem, namely that it is difficult to find _where_ a subroutine is even if you know its name. As it stands, this functionality could be improved because, e.g., you don't always know the _exact_ name of a subroutine, just a rough guess. This function also only returns (a list of) a list of children: `[child_a, child_b, child_c]` which means to retrieve the subroutine you want, you need to call:
```
possible_pathways: list[list[str]] = routine.find_descendants(name=subroutine_i_want)
possible_routines = []
for child_path in possible_pathways:
    routine_i_want = routine
    for child in child_path:
        routine_i_want = routine_i_want.children[child]
    possible_routines.append(routine_i_want)
```
So: if there are multiple children with the same name, `find_descendants` returns all pathways to those children. To get the actual child routine, you need to traverse the DAG  manually with `.children`, either in a `for` loop like above or chained together: `.children[...].children[...].children[...]`. 

We could then feasibly introduce functionality to streamline this for the user. 

I would also lump in [this issue](https://github.com/PsiQ/bartiq/issues/189) to be a similar vein. 

So we have two tracks: the under-the-hood behaviour of Bartiq as a compilation engine, and how users actually want to interact with `Routines/CompiledRoutines`. 

One proposal is to implement something like a `RoutineHandler` that allows `BaseRoutine` objects to be interacted with in a more interactive way, but I'm curious to hear peoples thoughts on this.

PR #235, opened 2025-06-24
refactor: modify SympyBackend and SympyInterpreter to allow functions to be overridden

Brendan-Reid1991:
## Description

Added a classmethod to `SympyBackend` to override the use of our custom `Max` function ([link](https://github.com/PsiQ/bartiq/blob/74ed648ddfb24002766340a45135192156791bee/src/bartiq/symbolics/sympy_interpreter.py#L182)). While the custom `Max` is a major source of our performance improvements during compilation, during the rewriter work the simplifications that the built-in Sympy `Max` function does are useful, and this is a simple way to allow users to access this functionality. 

I also added a docstring to the `SympyBackend` class to attempt to explain this, although most users interact with the already-instantiated `sympy_backend` (if they interact with it at all), so I'm wondering if there is a more visible way to convey this information. 

- Why isn't it just a kwarg?
- - As mentioned above, most users interact with the _variable_ `sympy_backend`, rather than importing the class. If it was a kwarg, the user would have to import `SymbolicBackend` and instantiate it themselves, _or_ another variable `sympy_backend_with_sympy_max` would need to be defined and exposed. And that just like, sucks, man. 


mstechly (changes requested):
I don't think this is the right approach to solve this problem.

1. This solution is very specific for sympy max. What if at some point we decide that we need a custom `min`?  Then we would have 4 different methods for each combination?
2. It only touches `as_expression` method and not other methods of the backend. For example, `max` method would stay the same, so if someone does `backend_with_sympy_max = backend.with_sympy_max()` and then `backend.max()`, they'll still use non-sympy max, which might lead to some very nasty errors.
3. It introduces some new elements into quite user-facing API, which I think can be avoided.

An idea for how to handle it better:

Introduce `overrides` argument to SympyBackend. This way you can do: `SympyBackend(overrides={"max": CustomMax})` and have an instance of a backend which does whatever you want. It's more scalable than adding attributes such as `_USE_SYMPY_MAX`.
Then, we would use different backend in compilation and in GSE. By default we would use one with overrides in compilation, and one without overrides (plain sympy functions) in GSE.
The only problem is the transition – i.e. when you get some expressions from compilation, they contain custom `Max`, so that needs to be replaced. However, we can do this with simple `backend.substitute`, so basically when you initialize GSE you would do: `expr = backend.substitute(expr, {CustomMax: SympyMax})`. 
I'm not a huge fan of hardcoding like this, but since we know there is only a single custom function we're using, I think it would be fine for now.

I think it also touches on similar things as https://github.com/PsiQ/bartiq/issues/228. 

And as a sidenote:
> As mentioned above, most users interact with the variable `sympy_backend`, rather than importing the class. 
I agree that this is status quo, but maybe this is not something we should be encouraging then? The sole fact that a certain user behavior exists, is not a convincing argument to me that it should stay this way.

src/bartiq/symbolics/sympy_backend.py
```diff
@@ -140,10 +142,56 @@ def _value_of(expr: Expr) -> Number | None:
...
+        ```python
+            backend_with_sympy_max = sympy_backend.with_sympy_max()
+            backend_with_sympy_max.as_expression("max(0, a)") == sympy_expr # True
+        ```
+
+    Args:
+        parse_function: A function that parses strings into Sympy expressions.
+    """
+
+    _USE_SYMPY_MAX: bool = False
```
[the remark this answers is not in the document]
Brendan-Reid1991: I don't really have a good reason, habit I think. I think I've always made class variables all caps because, although they aren't constant, they are "more constant" because they aren't instance variables. I can change this

Brendan-Reid1991:
just to make sure we're all on the same page... I'm happy defining another backed `compilation_efficient_sympy_backend` that uses the `CustomMax` function. But, I'll have to undo a lot of Konrad's changes in the PR that introduced the `CustomMax`. That is, we want to return `sympy_backend` (and therefore `SympyInterpreter`) to its original form and then modify it with the `CustomMax` when creating `compilation_efficient_sympy_backend`. If we didn't, we'd have this weird setup:
```
compilation_efficient_sympy_backend = SympyBackend()
sympy_backend = SympyBackend(overrides = {"max": sympy.Max})"
```
where the "default" sympy backend isn't the default at all. Not a breaking change, just a bit bigger of a change.

dexter2206 (approved):
Gert lush!

src/bartiq/symbolics/sympy_interpreter.py
```diff
@@ -258,10 +255,17 @@ def eval(cls, *args):
     "nlz": nlz,
 }
 
+SYMPY_ONLY = {"max": SympyMax}
+"""This dictionary lists all functions in `SPECIAL_FUNCS` we overrode with custom implementations."""
```
mstechly, line 259: [question] Does this appends this docstring to the `SYMPY_ONLY` variable so it's easier to access or sth? Cause normally I'd just do `# comment` above the definition.

src/bartiq/symbolics/sympy_backend.py
```diff
@@ -141,6 +143,12 @@ def _value_of(expr: Expr) -> Number | None:
 
 
 class SympyBackend:
+    """A backend for parsing symbolic expressions with Sympy.
```
mstechly, line 146: [nitpick] We usually put such a docstring into `init` and not under class. I think?

src/bartiq/analysis/_rewriters/sympy_rewriter.py
```diff
@@ -14,16 +14,21 @@
...
 from bartiq.analysis._rewriters.expression_rewriter import (
     ExpressionRewriter,
     ResourceRewriter,
     TExpr,
     update_expression,
 )
+from bartiq.symbolics.sympy_backend import SympyBackend, parse_to_sympy
+from bartiq.symbolics.sympy_interpreter import Max as CustomMax
+
+sympy_only_backend = SympyBackend(parse_function=partial(parse_to_sympy, built_in_sympy_only=True))
```
mstechly, line 31: [question] Seems to me like it would make more sense if we defined it in `sympy_backend` and imported from there?

src/bartiq/symbolics/sympy_interpreter.py
Brendan-Reid1991: `pylance` detects it as a docstring if you hover over it!

src/bartiq/symbolics/sympy_backend.py
Brendan-Reid1991: I was trying to find something in PEP about this, but I've never seen a docstring in an `__init__`! Happy to be proven wrong

src/bartiq/analysis/_rewriters/sympy_rewriter.py
Brendan-Reid1991: This has been removed!

src/bartiq/symbolics/sympy_backend.py
```diff
@@ -86,7 +84,11 @@ def _inner(backend: SympyBackend, expr: TExpr[S], *args: P.args, **kwargs: P.kwa
     return _inner
 
 
-def parse_to_sympy(expression: str, debug: bool = False) -> Expr:
+class Parser(Protocol):
+    def __call__(self, expression: str, *args, **kwargs) -> Expr: ...  # noqa: E704
```
Brendan-Reid1991, line 88: `black` is forcing this onto one line, but flake8 is complaining

src/bartiq/symbolics/sympy_backend.py
mstechly: Hmm... I found some examples of that in our internal repos, but actually it's not something we would typically do. 
Sorry!

mstechly (changes requested):
One thing that I see lacking is tests of `SympyBackend` with `use_sympy_max=True`. It's only tested as part of rewriter, but not as it's own thing in `sympy_backend` tests.

PR #238, opened 2025-07-04
Substitutions

Brendan-Reid1991:
## Description

A mammoth PR. Happy to split this into two or more if necessary. 

## Cliff Notes

- Added an ignore code to flake8. When defining abstract methods with `...`: `black` forces `...` commands onto the same line as the `def`, which flake8 hates. Conversely, having a `pass` statement instead of `...` raises this as dead code in pytest, and contributes to lower test coverage. I think adding this code to be ignored because all other instances of multiple statements on one line will be caught and corrected by black!
- Created and exposed the `sympy_rewriter` factory method in `bartiq/analysis/rewriters`. Also, made this folder public. 
- Rather than an `Instructions` Enum and trying to make `Assumptions` and `Substitutions` fit that format, I made a dummy class that all instructions inherit from, including `Assumptions`. This makes type checking nicer.
- Added `substitute` and `_substitute` methods to the base class, and defined `_substitute` in the `SympyExpressionRewriter`. Logic here is fairly gnarly!
- Updated the `focus` method such that linked parameters are included. i.e. for an expression with a variable `a`, if we substitute `a` with `b`, and try to `focus('a')`, it will show us all terms with `b`. This works for multiple levels of substitution. 
- Updated tests to reflect all these changes


src/bartiq/analysis/rewriters/sympy_rewriter.py
```diff
@@ -202,3 +256,47 @@ class SympyResourceRewriter(ResourceRewriter[Expr]):
...
+    Args:
+        expression: The top-level SymPy expression to process.
+        pattern: The pattern expression to match against subexpressions.
+        replacement: The expression to replace matches with (uses substitutions).
+
+    Returns:
+        The new expression with all matching subexpressions replaced.
+    """
+
+    if any(isinstance(expression, t) for t in [Symbol, Number]):
```
dexter2206, line 274: **suggestion**: You can pass two classes to `isinstance` instead of looping over them:

```suggestion
    if isinstance(expression, (Symbol, Number)):
```

src/bartiq/analysis/rewriters/sympy_rewriter.py
```diff
@@ -202,3 +256,47 @@ class SympyResourceRewriter(ResourceRewriter[Expr]):
...
+        if (matches := expression.match(pattern)) and not _all_values_numeric(matches.values())
+        else expression
+    )
+
+    if hasattr(replaced_expr, "args") and replaced_expr.args:
+        replaced_expr = replaced_expr.__class__(
+            *tuple(_replace_subexpression(arg, pattern, replacement) for arg in replaced_expr.args)
+        )
+
+    return replaced_expr
```
dexter2206, line 291: **suggestion**: This is a very common patter, where we want an attribute to exist and be nonempty (or, more generally, to be "truthy"). This can be achieved more concisely using `getattr`:

```suggestion
    if args := getattr(replaced_expr, "args"):
        replaced_expr = replaced_expr.__class__(
            *tuple(_replace_subexpression(arg, pattern, replacement) for arg in args)
        )

    return replaced_expr
```

src/bartiq/analysis/rewriters/utils.py
```diff
@@ -61,12 +85,47 @@ def from_string(cls, assumption_string: str) -> Self:
...
 
 
+@dataclass(frozen=True)
+class Substitution(Instruction):
+    """Substitute a symbol with an expression."""
+
+    symbol_or_expr: str
+    replacement: str
+    backend: SymbolicBackend
+    wild: list[str] = field(default_factory=list)
```
dexter2206, line 95: **suggestion**: Let's make it a tuple, it fits better the this frozen dataclass.

```suggestion
    wild: tuple[str, ...] = field(default_factory=tuple)
```

src/bartiq/analysis/rewriters/utils.py
```diff
@@ -61,12 +85,47 @@ def from_string(cls, assumption_string: str) -> Self:
...
+
+        return {
+            x: y
+            for x in free_symbols_in_expr(self.replacement)
+            if x not in (y := free_symbols_in_expr(self.symbol_or_expr))
+        }
+
+
+def _get_wild_characters(expression: str) -> list[str]:
+    return re.findall(r"\$([a-zA-Z_][a-zA-Z0-9_]*)", expression)
```
dexter2206, line 115: **suggestion**: As per above:

```suggestion
def _get_wild_characters(expression: str) -> tuple[str, ...]:
    return tuple(re.findall(r"\$([a-zA-Z_][a-zA-Z0-9_]*)", expression))
```

src/bartiq/analysis/rewriters/utils.py
```diff
@@ -85,10 +144,7 @@ def _get_properties(comparator: str, reference_value: int | float) -> dict[str,
     value_positive: bool = reference_value >= 0
     value_negative: bool = reference_value <= 0 or (not value_positive)
 
-    return {
-        "positive": ((gt or gte) and value_positive) or None,
-        "negative": ((lt or lte) and value_negative) or None,
-    }
+    return {"positive": ((gt or gte) and value_positive) or None, "negative": ((lt or lte) and value_negative) or None}
```
dexter2206, line 147: **issue**: I missed this in the previous review, but we don't handle the case of 0 correctly, or at least the naming is incorrect.

1. If `reference_value` is 0, both `value_negative` and `value_positive` are true.
2. Depending on the operator, either "positive" or "negative" is set to true, but if `gte` or `lte`, the symbol could be actually 0, which is neither positive nor negative.

Maybe we just should call those keys "nonnegative" and "nonpositive" instead of "positive" and "negative"?

src/bartiq/analysis/rewriters/sympy_rewriter.py
```diff
@@ -202,3 +256,47 @@ class SympyResourceRewriter(ResourceRewriter[Expr]):
...
+
+
+def sympy_rewriter(expression: str | Expr) -> SympyExpressionRewriter:
+    """Initialize a Sympy rewriter instance."""
+    match expression:
+        case str():
+            return SympyExpressionRewriter(_SYMPY_BACKEND.as_expression(expression))
+        case Expr():
+            return SympyExpressionRewriter(expression.replace(CustomMax, Max))
```
dexter2206, line 300: 1️⃣ 

```suggestion
        case str():
            return SympyExpressionRewriter(
                (expr := _SYMPY_BACKEND.as_expression(expression)),
                _original_expression=expr
            )
        case Expr():
            return SympyExpressionRewriter(
                (expr := expression.replace(CustomMax, Max)),
                _original_expression=expr
            )
```

dexter2206 (changes requested):

.flake8
```diff
@@ -26,6 +26,8 @@ ignore =
     # Docstrings for public modules and packages
     D100,
     D104,
+    # Ignore multiple statements on one line
+    E704,
 per-file-ignores =
     tests/*:D100,D101,D102,D103,D104
     __init__.py:F401
```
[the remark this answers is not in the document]
Brendan-Reid1991: There is one other use of this in `sympy_backend` ([link](https://github.com/PsiQ/bartiq/blob/7f69ef476152917b7196dda86a119e3fe28dd3f5/src/bartiq/symbolics/sympy_backend.py#L88)). I created that protocol during the refactor of the sympy backend, when we added arguments to the parser that broke some of the typing and this was the fix I found. 

But, I am adding a per-file-ignore for the lambda functions in `_wildcard_substitutions`.

src/bartiq/analysis/rewriters/sympy_rewriter.py
```diff
@@ -0,0 +1,302 @@
+# Copyright 2025 PsiQuantum, Corp.
```
Brendan-Reid1991, line 1: I made sure to `git mv` this file so this exact scenario didn't happen, but here we are.

src/bartiq/analysis/rewriters/expression_rewriter.py
```diff
@@ -134,45 +147,62 @@ def individual_terms(self) -> Iterable[T]:
...
             current = current.assume(assumption=assumption)
-        return replace(self, expression=current.expression, _previous=(Instruction.ReapplyAllAssumptions, self))
+        return replace(self, expression=current.expression, _previous=(ReapplyAllAssumptions(), self))
+
+    @abstractmethod
+    def _substitute(self, substitution: Substitution) -> T: ...
+
+    def substitute(self, symbol_or_expr: str, replace_with: str) -> Self:
+        """Substitute a symbol or subexpression for another symbol or subexpression.
+        By default performs a one-to-one mapping, unless wildcard symbols are implemented.
```
mstechly, line 191: [minor] "unless wildcard symbols are implemented. -> "unless there are wildcards present in `replace_with`" ? I'm not entirely sure what you meant here.

Also, it looks to me that this docstring (and some other as well) are missing `Args` section :)

src/bartiq/analysis/rewriters/utils.py
```diff
@@ -16,11 +16,16 @@
...
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from enum import Enum
+from typing import Iterable
 
 from typing_extensions import Self
 
+from bartiq.symbolics.backend import SymbolicBackend
+
+WILDCARD_FLAG = "$"
```
mstechly, line 27: [minor] I'd call it `WILDCARD_SYMBOL`, as when I see "flag" I think "boolean flag" and it's definitely not it. Also, we have wildcards already [in AST parser](https://github.com/PsiQ/bartiq/blob/7f69ef476152917b7196dda86a119e3fe28dd3f5/src/bartiq/symbolics/ast_parser.py#L123) and we use a `~` symbol. WDYT about changing this symbol to `$` in the parser to make it consistent?
I think `$` is a better symbol than `~` as the latter has some mathematical meaning, while the former doesn't.

src/bartiq/analysis/rewriters/utils.py
```diff
@@ -31,22 +36,41 @@ class Comparators(str, Enum):
...
 
-@dataclass
-class Assumption:
-    """A simple class for storing information about symbol assumptions."""
+class Instruction:
+    """Base class for all rewriter-transforming instructions."""
+
+
+@dataclass(frozen=True)
+class Initial(Instruction):
```
mstechly, line 44: [question] Why these aren't just enums?
Ok, I see later that we have some `__post_init__`, that's probably the reason?

src/bartiq/analysis/rewriters/utils.py
```diff
@@ -61,12 +85,47 @@ def from_string(cls, assumption_string: str) -> Self:
         return cls(*_unpack_assumption(assumption_string))
 
 
+@dataclass(frozen=True)
+class Substitution(Instruction):
+    """Substitute a symbol with an expression."""
```
mstechly, line 90: [minor] class attributes lack description. In particular `wild`, it's unclear to me what it represents.
Also, I personally wouldn't call it `symbol_or_expr`, but rather just `expr` – `symbol` is just a special case of a "single symbol expression", no?

src/bartiq/analysis/rewriters/utils.py
```diff
@@ -61,12 +85,47 @@ def from_string(cls, assumption_string: str) -> Self:
...
+    return tuple(re.findall(r"\$([a-zA-Z_][a-zA-Z0-9_]*)", expression))
+
+
 def _get_properties(comparator: str, reference_value: int | float) -> dict[str, bool | None]:
     """Derive properties of an assumption.
 
-    At present this only detects positivity/negativity.
-
-    If the properties are unknowable due to lack of information, they are None.
+    At present this only detects positivity/negativity. The two are calculated independently,
```
mstechly, line 121: [nitpick] `this only detects positivity/negativity` -> `this only detects positivity/negativity of a given symbol/expression` ?

src/bartiq/analysis/rewriters/utils.py
```diff
@@ -61,12 +85,47 @@ def from_string(cls, assumption_string: str) -> Self:
...
+
+    symbol_or_expr: str
+    replacement: str
+    backend: SymbolicBackend
+    wild: tuple[str, ...] = ()
+
+    def __post_init__(self):
+        object.__setattr__(self, "wild", _get_wild_characters(self.symbol_or_expr))
+
+    def _get_linked_parameters(self) -> dict[str, Iterable[str]]:
```
mstechly, line 100: [issue] I don't like this method, I think there are a couple of reasons:
1. It's a private method but it's used in `expression_rewriter`. Maybe it should be public then?
2. It could be just a function, which takes in `Substitution` object, doesn't need to be a class method. It's the only `Instruction` which has some special methods defined (except `from_str` for `Assumption`, but that's 100% sensible to me), so it breaks symmetry a bit.

Thoughts?

src/bartiq/analysis/rewriters/utils.py
mstechly: I'm fine either way.  I think `nonnegative` and `nonpositive` are natural words to use when you're mathematicians, otherwise they take a bit to parse. 
I think that this is a place in the code where the choice will not be very consequential.

tests/analysis/rewriters/basic_rewriter_tests.py
```diff
@@ -101,10 +110,10 @@ def test_history(self):
...
         updated_rewriter = init_rewriter.expand().simplify().assume("beth>0")
         assert updated_rewriter.history() == [
-            Instruction.Initial,
-            Instruction.Expand,
-            Instruction.Simplify,
-            "Assumption(beth>0)",
+            Initial(),
+            Expand(),
+            Simplify(),
+            Assumption.from_string("beth>0"),
```
mstechly, line 116: [question] Actually, why it's not  just`Assumption("beth>0")`?

mstechly (changes requested):

.flake8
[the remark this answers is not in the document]
dexter2206: Putting noqa in specific file would not work (or maybe I was doing something wrong), so I think we are down to either local noqas or this per-file-ignores in the config. I think per-file-ignores makes sense *in this particular case*, because TBH how this rule is implemented is messed up. There is no way to exclude ellipsis on the same line (which is a common formatting practice for protocol methods), and ignoring each one of those separately seems messy.

src/bartiq/analysis/rewriters/expression_rewriter.py
```diff
@@ -18,67 +18,82 @@
...
     @property
     def original(self) -> Self:
         """Return a rewriter with the original expression, and no modifications."""
-        return type(self)(expression=self._original_expression, backend=self.backend)
+        return type(self)(
+            expression=self._original_expression, _original_expression=self._original_expression, backend=self.backend
+        )
 
-    def _unwrap_history(self) -> list[tuple[Instruction | str, ExpressionRewriter[T] | None]]:
+    def _unwrap_history(self) -> list[tuple[Instruction, ExpressionRewriter[T] | None]]:
```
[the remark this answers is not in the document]
dexter2206: I'll answer this because it was me who suggested immutable rewriters, for which the current implementation of history is the only one that makes sense.

Storing of the history is almost free in terms of both time and memory, because each new rewriter instance stores only one existing instance (namely: the one it originated from). Therefore, comparing to not having a history, the cost is roughly around 1 variable assignment + passing variable to the initializer, which is miniscule.

Retrieving history is of course linear in the number of history items. A quick benchmark that I just performed  shows that obtaining a history of a rewriter after a 1000 steps takes around 61 microseconds on my M1. In the benchmark, I created an expression summing variables x_1 + ... x_1000, and then substituted x_i -> y_i one by one (to get this nice 1000-items long history). Also, I measured the memory footprint of a rewriter with such history, and the whole thing was talking around 1.7MB.

**Nerding out on the impact of storing the history.**
To properly isolate how storing a history in this way impacts overall performance, we canperform a simple benchmark with two classes differing the same operation, except one stores the history and one does not.

```python
@dataclass
class Foo:
    x: int
    previous: Foo | None = None
    def next(self):
        return replace(self, x=self.x+1, previous=self)

@dataclass
class Bar:
    x: int
    def next(self):
        return replace(self, x=self.x+1)

def f(cls, n_iters):
    obj = cls(1)
    for _ in range(n_iters):
        obj = obj.next()
```

Measuring `f(Foo, n)` and `f(Bar, n)` on my machine shows that the difference in performance is roughly 0.7 microseconds per 10 iterations, which is negligible.

tests/analysis/rewriters/basic_rewriter_tests.py
Brendan-Reid1991: I tried to get that format working a couple of different times, and while it _is_ possible, the logic is very messy. 

Unless we want to move to only allowing strings as input, i.e.:
```python
@dataclass(frozen=True)
class Assumption(Instruction):
    assumption_string: str
```

in order to have the current implementation also accept strings as well as the three current arguments, we'd have to overwrite `__new__` to handle strings only, and fallback to the __init__ for everything else. And because it's a dataclass, it's annoying.

 I did this a few weeks ago when I was writing the Assumptions, and when I finally got it working the class looked like a complete mess. I decided it wasn't worth it, because users most likely won't interact with `Assumption` directly anyway.

src/bartiq/analysis/rewriters/utils.py
Brendan-Reid1991: Possibly a nice middle ground:
- Made it a private function that takes in a `Substitution`
- Added a new input to the `Substitution` dataclass that has `init=False`
- set the attr in the `_post_init_`
- Logic is outside `Substitution`, but the attribute is kept there. i.e function is no longer needed in `expression_rewriter`

.flake8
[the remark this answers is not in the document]
mstechly: Ok, sounds good to me!

tests/analysis/rewriters/basic_rewriter_tests.py
mstechly: Ok, thank you for the clarification!

PR #240, opened 2025-07-09
fix: make ntz(0) = -1

dexter2206:
## Description

This PR makes `ntz(0)` equal to -1, as discussed via other communication channels.

Also cleans up relevant test cases.


mstechly (approved):
Please just add information about the convention we're using in `ntz` docstring.

PR #241, opened 2025-07-15
chore: docstring updates and housekeeping on rewriters

Brendan-Reid1991:
## Description

After a sync with @mstechly yesterday we identified some minor improvements that can be made to the new rewriter interface. 

## Overview of changes:
- Exposed `ResourceRewriter` at the `bartiq/analysis` level
- Docstring improvements
- Added a `from_history` class method to port instructions over to different resources in a routine.
- Made fields in `Assumptions` and `Substitutions` properties instead.
- Renamed linked_params(eters) -> linked_symbols
- Added a test for `from_history`

## Changes desired but not implemented

At the moment the `ResourceRewriter` does not contain any typehinting or auto-completion for methods that it inherits from the `ExpressionRewriter`.  I wrote a protocol that would at least autocomplete some of the methods, but the protocol interferes with the `__getattr__` functionality and stops it working. An easy alternative is to manually forward all the attribute calls, but this causes a decent bit of boilerplate. ChatGPT suggested a bespoke meta class, so I might give that a try. 


src/bartiq/analysis/rewriters/expression.py
```diff
@@ -36,12 +36,24 @@
...
 class ExpressionRewriter(ABC, Generic[T]):
-    """An abstract base class for rewriting expressions."""
+    """An abstract base class for rewriting expressions.
+
+    Args:
+        expression: The expression of interest.
+        _original_expression: Private attribute to store the original expression with no modifications.
+            On initialisation, should be set to equal self.expression.
+        backend: Which symbolic backend to use.
+        linked_symbols: A dictionary of symbols that are linked, i.e. which have been substituted for which. Helps
```
mstechly, line 46: "which have been substituted for which" ?

src/bartiq/analysis/rewriters/expression.py
```diff
@@ -36,12 +36,24 @@
 
 @dataclass
 class ExpressionRewriter(ABC, Generic[T]):
-    """An abstract base class for rewriting expressions."""
+    """An abstract base class for rewriting expressions.
+
+    Args:
```
mstechly, line 41: I think this would be `Attributes ` rather than `Args` for a class?

src/bartiq/analysis/rewriters/utils.py
```diff
@@ -77,10 +77,13 @@ class Assumption(Instruction):
...
     value: int | float
-    symbol_properties: dict[str, bool | None] = field(init=False)
 
-    def __post_init__(self):
-        object.__setattr__(self, "symbol_properties", _get_assumption_properties(self.comparator, self.value))
+    @property
+    def symbol_properties(self) -> dict[str, bool | None]:
+        """A dictionary of properties (positive, negative, non-positive, non-negative).
+
+        If None, the truthiness of the property could not be determined."""
```
mstechly, line 85: None -> `None` .
Also, perhaps rather than `truthiness` we could say "the boolean value of the property ..." ?

src/bartiq/analysis/rewriters/utils.py
```diff
@@ -103,33 +106,56 @@ class Substitution(Instruction):
...
-        wild: A tuple of symbols that have been marked as "wild", i.e. they match anything.
+
+    Attributes:
+        wild: A tuple of symbols that have been flagged as 'wild' in the expr.
+        linked_symbols: A dictionary of which symbols have been replaced with which, e.g.
+            ```python
+                sub = Substitution(expr="a+b", replacement="c", backend=sympy_backend)
+                sub.linked_symbols # {"c": ("a", "b")}
+            ```
+            If any symbols are 'wild', linked_symbols is an empty dictionary as this corresponds to a many-to-many
```
mstechly, line 117: linked_symbols -> `linked_symbols`

mstechly (approved):

PR #244, opened 2025-07-24
chore: documentation for rewriters

Brendan-Reid1991:
## Description

I wrote a first draft of the documentation for rewriters. Please read and if we don't manage to make all the necessary changes before Friday evening, I am happy for someone else to take over and finish it in my absence. 

At one point I attempted to link directly to the API reference , and found that the table of contents in `reference.md` was broken. I added some title lines to `reference.md`, as well as
```yaml
- toc:
    permalink: true 
```
to `mkdocs.yaml` to allow for linking (I didn't actually do this in the end, but it seems like useful functionality)

and I changed `show_submodules: true->false`, and that's what seemed to fix to broken table of contents. 

### DIFF
--- a/docs/concepts/rewriters.md
+++ b/docs/concepts/rewriters.md
@@ -5,78 +5,561 @@
 ```python
 from bartiq.analysis import sympy_rewriter
 ```
-Here we will explain the functionality common to all backend rewriters, and the functionality specific to the SymPy backend.
+The `sympy_rewriter` operates on a single expression, however we also provide a class object to operate on `bartiq` `CompiledRoutine` objects:
+```python
+from bartiq.analysis import ResourceRewriter
+```
+Here we will give an overview of the functionality currently implemented in these objects, with example usage.
 
 ## Motivation
 
-As quantum algorithms increase in complexity their symbolic resource expressions similarly become more complex. For a state of the art algorithm like double factorization (link) the resource expressions can be almost impossible to parse, due to the sheer number of terms and symbols. Making complex expressions more palatable is a primary motivation for rewriters, however we also want to implement standard simplifications. Terms like `max(0, x)` can be simplified if we could just indicate that `x>0`. 
+As quantum algorithms increase in complexity their symbolic resource expressions similarly become more complex. For a state of the art algorithm like double factorization (find a link to put here) the resource expressions can be almost impossible to parse, due to the sheer number of terms and symbols. Making complex expressions more palatable is a primary motivation for rewriters; gaining insights into closed-form expressions for important resource quantities is vital for fault-tolerant quantum algorithm optimization and design.
 
 ## Overview 
 
-Rewriters are structured as dataclasses with associated methods and properties. The input should be an expression that we wish to modify (or, rewrite). While much of the logic is necessarily tied to a particular backend implementation, the base class enforces some core functionality. The philosophy around rewriters is that they should be _immutable_
+Rewriters are structured as dataclasses with associated methods and properties. Instantiation is done through a factory method; the only required input is an expression that we wish to modify (or rewrite). The input can be provided as a string, or as the backend-specific expression type.
+
+While much of the logic is necessarily tied to a particular backend implementation, the base class enforces some core functionality. The philosophy around rewriters is that they should be _immutable_, such that methods that change an expression actually return a new instance of the rewriter class. This allows for method chaining and easy access to previous expression forms if a change was made in error.
+
+Due to the dynamic nature of expression manipulation, rewriters are designed with interactive environments in mind. Rewriters have a `_repr_latex_` method that prints the current expression, meaning the following code in a Jupyter notebook:
+```python
+sympy_rewriter("a + b")
+```
+would return a $\LaTeX$ (technically $KaTeX$) expression $a + b$. This, combined with method chaining, means the effect of different methods can be seen immediately.
 
 
+## Concepts
+In designing the rewriter framework we implemented a number of different utility classes. A typical user should not need to interact with these objects, but we describe them here for completeness.
 
+#### Instructions
+An `Instruction` is an action that marks a change to an expression. The following `Instructions` are implemented:
+
+- `Initial`
+
+    The initial form of the rewriter instance.
+
+- `Simplify`
+
+    A backend-specific 'simplify' command.
+
+- `Expand`
+
+    Expand all brackets in the expression.
+
+- `Assumption`
+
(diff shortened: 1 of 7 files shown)


docs/concepts/rewriters.md
```diff
@@ -0,0 +1,565 @@
...
+```
+The `sympy_rewriter` operates on a single expression, however we also provide a class object to operate on `bartiq` `CompiledRoutine` objects:
+```python
+from bartiq.analysis import ResourceRewriter
+```
+Here we will give an overview of the functionality currently implemented in these objects, with example usage.
+
+## Motivation
+
+As quantum algorithms increase in complexity their symbolic resource expressions similarly become more complex. For a state of the art algorithm like double factorization (find a link to put here) the resource expressions can be almost impossible to parse, due to the sheer number of terms and symbols. Making complex expressions more palatable is a primary motivation for rewriters; gaining insights into closed-form expressions for important resource quantities is vital for fault-tolerant quantum algorithm optimization and design.
```
mstechly, line 16: (find a link to put here) Ha!
I think Fig 16 from https://arxiv.org/abs/2011.03494 is a good reference to put.

docs/concepts/rewriters.md
```diff
@@ -0,0 +1,565 @@
...
+    Substitute a symbol or expression for another.
+
+- `ReapplyAllAssumptions`
+
+    Reapply all previously applied assumptions.
+
+The primary purpose of `Instructions` is to track the history of an expression, and the user may only ever interact with these objects via the `history()` method. Of these classes, only `Assumption` and `Substitution` require special logic; the others have no methods or attributes implemented. We explore `Assumption` and `Substitution` in more detail below.
+
+!!! note "Why not just an `Enum`?"  
+    Originally the `Instructions` were implemented as an `Enum`! However, because `Assumptions` and `Substitutions` required special logic it was challenging to enforce strict typing across both an `Enum` and other dataclasses. Creating an empty class `Instruction` that everything else subclassed from is a cleaner implementation.
```
mstechly, line 64: [nitpick] I'd say "resulted in a cleaner implementaiton".

docs/concepts/rewriters.md
```diff
@@ -0,0 +1,565 @@
...
+
+Alternatively an assumption can be parsed directly from a string:
+```python
+sympy_rewriter("max(0, X)").assume("X > 0")
+>>> X
+```
+
+Given an input assumption to a `sympy_rewriter`, the symbol is updated with the relevant SymPy [predicates](https://docs.sympy.org/latest/guides/assumptions.html#predicates). For some symbol `X` the predicates we support are:
+
+- positive: `X > a > 0` or `X > 0`,
```
mstechly, line 90: I don't get what's the difference between `X > a > 0` and `X > 0` in this context.

docs/concepts/rewriters.md
```diff
@@ -0,0 +1,565 @@
...
+- nonnegative: `X >= 0` or `positive`,
+- negative: `X < a < 0` or `X < 0`,
+- nonpositive: `X<=0` or `negative`,
+
+where `a` is some real number. From these SymPy is able to deduce other predicates. We do not implement predicates that declare if a symbol belongs to a particular number group, i.e. `integer`, `complex`, etc. We found that these kinds of assumptions did little to help simplify expressions. Similarly we do not have support for symbols being declared as infinitely large; if there is a valuable use case for these then they are easily added.
```
mstechly, line 95: "they could be easily added"

docs/concepts/rewriters.md
```diff
@@ -0,0 +1,565 @@
...
+
+
+There is unfortunately [no way to input assumptions between different symbols](https://docs.sympy.org/latest/guides/assumptions.html#relations-between-different-symbols). Similarly SymPy does not implement predicates that specify a relationship between a symbol and some nonzero value, i.e. `X > 5`. However, for this latter point, we have implemented a workaround.
```
mstechly, line 98: It's great that we have a a very specific link to sympy docs which show that it's not possible. Do we also give people this link when we throw a warning in that method? Could be a good idea.

docs/concepts/rewriters.md
```diff
@@ -0,0 +1,565 @@
...
+Finally, it is possible to mix-and-match wild symbols with non-wild symbols:
+```python
+sympy_rewriter(
+    "a*max(0, x) + b*max(0,y) + a*max(0, y)"
+).substitute("a*max(0, $x)", "a*x")
+>>> a*x + a*y + b*max(0, y)
+```
+
+##### Caveats
+Here we collect some caveats for wildcard substitutions.
```
mstechly, line 206: Wasn't there sth about having 0 as an argument of a function as well? Maybe it's elsewhere though 🤷‍♂️

docs/concepts/rewriters.md
```diff
@@ -0,0 +1,565 @@
...
+- `expression: T`
+
+    The form of the current expression. This is the attribute updated by rewriting methods and displayed by the `_repr_latex_` method in Jupyter notebooks.
+
+    ```python
+    sympy_rewriter("a + b + c").expression
+    >>> a + b + c
+    ```
+
+- `linked_symbols: dict[str, Iterable[str]]`
```
mstechly, line 303: Haven't we decided to change the name in the end?

docs/concepts/rewriters.md
```diff
@@ -0,0 +1,565 @@
...
+
+- `routine: CompiledRoutine`
+- `resource: str`
+- `history: list[Instruction]`
+- `rewriter_factory: ExpressionRewriterFactory = sympy_rewriter`
+
+and returns an instance of `ResourceRewriter` with the history loaded in. 
+
+## Implementation details
+Below we list some of the most important attributes, properties and methods. In what follows, the typehint `T` is used to indicate that the type is backend-dependent expression type. For instance in the `sympy_backend`, `T = sympy.Expr`.
```
mstechly, line 289: and methods... of what?
Since this is directly after `ResourceRewriter`, it feels like this section is only about it. I think it's worth making it clear that this is about `Rewriter`.

mstechly: Also, I think that it might not be worth having this session, as this info should be included in the API docs and be more up to date, so maybe we should just link to the appropriate place in the API docs? WDYT?

mstechly (changes requested):
API docs seem incomplete to me.
Screenshot 2025-07-24 at 4 21 17 PM

pqvr (approved):
LGTM! 🚀

Issue #247, opened 2025-08-04
Rewriter fails if initialized with integer


mstechly:
## Describe the bug
When dealing with a dictionary of resources, some of the values might be integers and not expressions.
In such case rewriter will throw an error

## How to Reproduce

```
from bartiq.analysis import sympy_rewriter
sympy_rewriter(0)
```

## Expected behavior
I think that it should also accept numbers.

## Actual behavior
We're getting `ValueError` with: `Invalid input type`
CC: @Brendan-Reid1991

Brendan-Reid1991:
Resolved by [#249](https://github.com/PsiQ/bartiq/pull/249)

PR #249, opened 2025-08-11
fix: allow numbers to be passed into sympy_rewriter

Brendan-Reid1991:
## Description

To address #247.

Added a test to ensure the `Max` function is correctly replaced as well.


mstechly (approved):
I'd also like to see a test which checks that it works with numbers :)

